PR #64, opened 2018-11-29
Minor appendix corrections.

vishaalkapoor:
- Use $n$th instead of $n$-th to be consistent with rest of the book.
- Correct language in notation section.
- Revert \textit.

### DIFF
--- a/chapter_appendix/math.md
+++ b/chapter_appendix/math.md
@@ -172,7 +172,7 @@ Here, $x_{ij}$ is the element of matrix $\boldsymbol{X}$ in row $i$ and column $
 ### Eigenvectors and Eigenvalues
 
 
-Let $\boldsymbol{A}$ be a matrix with $m$ rows and $n$ columns. If $\lambda$ is a scalar and $\boldsymbol{v}$ is a non-zero $n$-dimensional vector with
+Let $\boldsymbol{A}$ be a matrix with $n$ rows and $n$ columns. If $\lambda$ is a scalar and $\boldsymbol{v}$ is a non-zero $n$-dimensional vector with
 
 $$\boldsymbol{A} \boldsymbol{v} = \lambda \boldsymbol{v},$$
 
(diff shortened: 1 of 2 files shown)


chapter_appendix/math.md
```diff
@@ -220,7 +220,7 @@ The Taylor expansion of function $f$ is given by the infinite sum
 
 $$f(x) = \sum_{n=0}^\infty \frac{f^{(n)}(a)}{n!} (x-a)^n,$$
 
-when it exists. Here, $f^{(n)}$ is the $n$-th derivative of $f$, and $n!$ is the factorial of $n$. For a sufficiently small number $\epsilon$, we can replace $x$ and $a$ with $x+\epsilon$ and $x$ respectively to obtain
+when it exists. Here, $f^{(n)}$ is the $n$th derivative of $f$, and $n!$ is the factorial of $n$. For a sufficiently small number $\epsilon$, we can replace $x$ and $a$ with $x+\epsilon$ and $x$ respectively to obtain
```
astonzhang: it may look clearer with -

chapter_appendix/math.md
```diff
@@ -232,7 +232,7 @@ $$f(x + \epsilon) \approx f(x) + f'(x) \epsilon.$$
 
 ### Partial Derivatives
 
-Let $u = f(x_1, x_2, \ldots, x_n)$ be a function with $n$ arguments. The partial derivative of $u$ with respect to its $i$-th  parameter $x_i$ is
+Let $u = f(x_1, x_2, \ldots, x_n)$ be a function with $n$ arguments. The partial derivative of $u$ with respect to its $i$th  parameter $x_i$ is
```
astonzhang: same here

chapter_appendix/math.md
vishaalkapoor: In the previous PR on the math appendix, I had changed this to $n$-th from $n$th for the reason of clarity, but then noticed there are 6 other occurrences in the text that use $x$th so I was changing it back for consistency.

I also did some searching on Stackoverflow for what seems to be the standard in mathematics texts and found that $n$th is in fact more widely used. 

Here's the stack overflow link: https://tex.stackexchange.com/questions/10959/1st-2nd-and-i-th

Let me know what you prefer

astonzhang: @smolix Alex, do you think we should go with $n$-th or $n$th? Alternatively, we can do $n^{\text{th}}$

PR #132, opened 2019-01-22
Update requests to 2.20.0

muelleme:
Otherwise conda env create -f environment.yml fails with:

mxnet 1.5.0b20181215 has requirement requests>=2.20.0,
but you'll have requests 2.18.4 which is incompatible.


astonzhang:
Thanks. I'll remove it.

PR #133, opened 2019-01-23
Fix wording, notation, and some math parts.

muelleme:
This mainly fixes wording, notational inconsistencies and some small typos in formulas in some places. More explanations can be found in the commit messages.

The problem is, I was not able to test these changes by building the book.
I fought a bit with the Makefile, but I think there are some files missing from the repository. At least I couldn't build and look at the generated html, which at least for the changes in the LaTeX parts should be done. I think it would be nice if you could add some instructions in the "how to contribute" section.

Another thing that I noticed but couldn't really fix without building and testing:
Line 110 in linear-algebra.md gets rendered as 
"But if we say :math:`n`-dimensional vector, we mean a vector of length :math:`n`." 
in the book. Something is wrong here with LaTeX-math-mode and the markdown-emphasis.

Also the code example after that sentence does not seem to be really related to that section...

### DIFF
--- a/chapter_crashcourse/install.md
+++ b/chapter_crashcourse/install.md
@@ -12,12 +12,12 @@ To get started we need to download and install the code needed to run the notebo
 
 For simplicity we recommend [conda](https://conda.io), a popular Python package manager to install all libraries.
 
-1. Download and install [Miniconda](https://conda.io/miniconda.html) at [conda.io/miniconda.html](https://conda.io/miniconda.html) based on your operating system.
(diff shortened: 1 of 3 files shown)


astonzhang:
If you want to produce html, do:

conda env create -f build/env.yml
conda activate d2l-en-build 
make html EVAL=0

Then you can view all the generated htmls in build/_build/html.

Note that the above doesn't generate code output. If you want to generate code output, remove EVAL=0. It will take about 2 hours if you have 2 gpus.

PR #145, opened 2019-02-03
Update rmsprop.md

FX196:
Same issue as in Chinese version, loading data two times


astonzhang:
Can you remove its second appearance (instead of 1st) on the same section? Thanks

PR #153, opened 2019-02-12
Minor revisions

simoncorstonoliver:
>> words are the basic unit of linguistic meaning
No, the basic unit of meaning in most frameworks for linguistic analysis is the morpheme. But no need to burden the reader with linguistic terminology.

I think the discussion on this page could motivate word2vec better. The primary issue with using one-hot encoding for words is data sparsity. Each word is its own idiosyncratic bundle of meaning, collocation and selectional restrictions but intuitively it feels like we could do a better job of representing less common words if we could generalize over sets of similar words.

Other ways to solve this problem of sparsity historically (before word2vec) include stemming and the use of part of speech tags.

The downside of using word2vec is that you're still not really solving the problem of morphology. This is a minor problem for the analysis of English and Chinese but more significant for many other languages.


astonzhang:
@simoncorstonoliver 

Hi Simon,

I hope that you are doing well.

Thanks for your contributions to d2l-en (https://github.com/d2l-ai/d2l-en). Do you mind re-licensing all your d2l-en contributions to us under the new license (MIT-0 for code and CC-BY-SA for text)? This will allow more people to use d2l-en for free. All you need to do is replying to this message. Thanks.

Regards,
Aston

PR #163, opened 2019-02-22
 Correct a statement in sampling.md

trungha-ngx:
> "The Normal distribution... is given by $p(x) = \frac{1}{\sqrt{2 \pi}} \exp\left(-\frac{1}{2} x^2\right)$."

-> 

> "The **standard** Normal distribution... is given by $p(x) = \frac{1}{\sqrt{2 \pi}} \exp\left(-\frac{1}{2} x^2\right)$."


trungha-ngx:
I made some mistakes when committing, so it looks odd now. Sorry about that.

PR #164, opened 2019-03-01
Missing than

mohamed-ali:


astonzhang:
Hi @mohamed-ali, would you like to send your name and GitHub ID, and mailing address to d2lbook.en@gmail.com? We'd like to mail you a hardcopy of the book when it's published. Thanks

PR #167, opened 2019-03-05
Fix the condition of vector derivative formulas in math.md

trungha-ngx:
For all the identities hold, **A** must be a square matrix of order n.

> "If **A** is a matrix with m rows and n columns, and **x** is an n-dimensional vector, the following identities hold..."

 -> 

> "If **A** is a square matrix of order n,..."


astonzhang:
Thanks. I think the original contents are correct.

astonzhang:
Closing this PR due to lack of activities :)

PR #172, opened 2019-03-09
Correct links to deep learning computation chapter.

vishaalkapoor:
Three links currently point to deep-learning-computation which seems to be an old path name for the Deep Learning Computation chapter. The current one is chapter_gluon-fundamentals. You can verify the links don't work. For example, 

https://d2l.ai/chapter_convolutional-neural-networks/lenet.html#LeNet contains the link
> https://d2l.ai/chapter_deep-learning-computation/use-gpu.md

This change will correct those links.

### DIFF
--- a/chapter_computational-performance/async-computation.md
+++ b/chapter_computational-performance/async-computation.md
@@ -156,7 +156,7 @@ def get_mem():
     return int(str(res).split()[15]) / 1e3
 ```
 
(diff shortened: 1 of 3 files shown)


astonzhang:
Thanks. We'll change the name of the directory back to  `chapter_deep-learning-computation`

vishaalkapoor:
That makes more sense. I'll remove this PR.

PR #179, opened 2019-03-15
Fix pReLU formula

vfdev-5:
Please, correct me if I'm wrong, but the formula for pReLU seems to be a bit different:
$$
pReLU(x) = \max(0, x) + \alpha \min(0, x)
$$
instead of currently
$$
pReLU(x) = \max(0, x) - \alpha x
$$

He et al 2015
Screen Shot 2019-03-15 at 10 32 52


smolix:
Good catch. In practice both things are equivalent up to scaling factors since you could always rescale it by $1/(1-\alpha)$. That said, it's probably best to keep the same notation as in the original pReLU paper. Thanks for your help!

PR #207, opened 2019-04-04
Fix typo in linear-regression.md

arush15june:


zackchase:
Great catches, thanks!!

PR #214, opened 2019-04-09
fixed test set size while generating data sets

prasanth5reddy:
In the chapter_multilayer-perceptrons/underfit-overfit.md it is mentioned that we will synthesize 100 samples each for the training set and test set. But in the code test set size is set to 1000 instead of 100. Hence fixed the test set size to 100.


zackchase:
Looks good.

PR #222, opened 2019-04-13
Updated some figure numbers.

brianhendee:
It looks like some additional figure numbers were off by 1. I did a search by chapter and updated.


mli:
Thanks for updating these numbers. I'm going to merge this PR. #221 is adding a new chapter 7 that cause all late figure numbers will be increased by 1. Can you please also fix them after #221 is merged?

PR #228, opened 2019-04-15
big revision to why-conv

zackchase:
big revision to why-conv...


zackchase:
big edit to conv section 1 (why-conv)

PR #234, opened 2019-04-17
Cnn5

zackchase:
edits to pooling and lenet


mli:
A new chapter will cause all figure numbers broken and links to the original pages, e.g. the course website, broken as well

zackchase:
@mli i know hut we've talked about this. we need to be able to organize at chapter level and fox this process. its inacceptable to have a book with 100 page “chapters”

PR #237, opened 2019-04-17
added loss from gluon in weight-decay.md

prasanth5reddy:
In the scratch implementation, both net and loss were defined from d2l package. While in gluon implementation loss still points to d2l package. This change has reassigned the loss to take from gluon in order to maintain concise implementation.


zackchase:
thanks!

PR #242, opened 2019-04-18
make it clearer

lkevinzc:
original is: decide when to read data into the cell

'when' and 'read' are confusing for input gate;
thus I change it to: decide 'how much data' to 'write' into the cell _for this time step_


astonzhang:
I would prefer the original one, just as an analogy of logic gates of a computer.

smolix:
@lkevinzc I would like to keep the original formulation. If you have a Boolean gate you either read or you don't read. Hence there isn't much in terms of the quantity of information. The input gate is either on or off. It's only when we move to differentiable continuous-valued models that this matters. 

Nonetheless, many thanks for your suggestion. Please keep up the good work.

PR #243, opened 2019-04-18
Typos and punctuation

Fietel:
Hey, 
went through 3 files, additional notes:
`chapter_appendix/notation.md` seems to have a problem during compilation, math code itself should be working (at Derivatives and Gradient)

`chapter_appendix/math.md`: Line 169 (Here, $x_{ij}$ is the element of matrix $\boldsymbol{X}$ in row $i$ and column $j$.) is a repetition of Line 41. Deletion could be considered

`chapter_appendix/how-to-contribute.md`:
1. Id suggest a consistent use of GitHub "ID" or "username" (preferably "username" imo)
2. Id also suggest not mixing up snake case and lower camel case for `your_Github_ID` (following py style: `your_github_id` or respectively `your_github_username`). If you prefer the mixing, the current version should be changed to `your_GitHub_ID` or `your_Github_Id`.
3. The example `~/repo/d2l-en/chapter_deep-learning-basics/linear-regression.md` does not utilize an existing folder due to past changes which might be confusing for some. I could provide an update+ updated screenshots if you want

Hope that this helped,
Fiete

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 This open source book has benefited from pedagogical suggestions, typo corrections, and other improvements from community contributors. Your help is valuable for making the book better for everyone. We will [acknowledge](https://www.d2l.ai/chapter_preface/preface.html#Acknowledgments) each D2L contributor in the book and send a free book (hard copy) to each contributor when it is published. 
 
-**Dear [D2L contributors](https://github.com/d2l-ai/d2l-en/graphs/contributors), please email your GitHub ID, name, and mailing address to d2lbook.en@gmail.com. Thanks.**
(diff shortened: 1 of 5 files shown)


astonzhang:
Thanks!

Would you like to make it consistent in using "username" rather than "ID"?

Fietel:
Sure, I just added two commits: one changing the "ID" to "username" and then fixing the mix of snake and lower camel case mentioned before (as a seperate commit in case you would like to keep the existing notation)

Fietel:
Quick question:
Some resources have changed and will continously change up until release (e.g. examples with chapters). Wouldn't it be useful to collect those passages that have to be checked again and corrected later on (instead of changing them repeatedly) and maybe leave a comment with a TODO or collect them in a single thread in the forum to correct them upon completion?

astonzhang:
Thanks! We are still editing the book. Hopefully we can (almost) finalize the first few chapters asap.

PR #249, opened 2019-04-20
Rephrased the introduction to AWS

Fietel:
Every local machine has limited computing resources (resources are limited by definition :D).
The relevant factor will probably be that they are not sufficient to perform a task in a specific amount of time.


Fietel:
There seems to be an issue with the formating. Double quotes in the .md are not correctly transformed to double quotes in the pdf (probably .md to the pdf via latex?). 
An example can be found in `chapter_appendix/aws.md` in line 22 with "Launch Instance". The double quotes dont appear in the pdf. Is it possible to simply use the `''` syntax of LaTeX?

astonzhang:
Thanks @Fietel and Alex!

PR #260, opened 2019-04-24
Review of chapter_install

Fietel:
1. Updated `apt` to `apt-get` according to the previous PR.
2. Reviewed `chapter_install`. Some reformulation, consistency updates and formatting.

Edit: 
Also the phrase "After the installation is complete, conda should be made to take effect..." sounds quite weird :)


zackchase:
Thanks for these updates! There's just a couple small mistakes. Please put back "both" into the two lists where it is appropriate and change "exemplary" back to "exemplar".

Fietel:
Updated "exemplar". Are you sure about "both"? It included more than two items in both of the lines I have changed

smolix:
> Updated "exemplar". Are you sure about "both"? It included more than two items in both of the lines I have changed and should be correct the way I updated it.
> 
> Edit:
> Issues with _both_ seem to be more common. Later on in `chapter_preface`:
> _[...]each subsequent notebook introduces both a reasonable number of new concepts and provides a single self-contained working example -- using a real dataset._
> The _both_ should be incorrect because in this case we _introduce_ one thing (a reasonable number of new concepts) and _provide_ another.
> Currently working on another branch to avoid too long PRs and I didnt yet delete the _both_ but would like to get some feedback :)

@Fietel - Zack is native speaker. I would trust his English more than anyone else's. Here's what you're probably getting confused about with regard to 'both ... and ...'. 
https://www.macmillandictionary.com/us/dictionary/american/both-and

Fietel:
Well that was not the problem, but using it as a suppletive variant would mean that you could replace it in a form of 
[...]each subsequent notebook _introduces both_ a reasonable number of new concepts and _introduces_ provides a single[..]
which is wrong. It is also another usecase as the ones mentioned by Zack (which e.g. was "_accessible **both** as executable code, as a physical book, as a downloadable PDF, and on the internet as a website._")
Using several hours in an attempt to help and receive such answers is quite demotivating :)

PR #295, opened 2019-05-16
fixed , to make reshape clearer in conv-layer.md

prasanth5reddy:


chapter_convolutional-neural-networks/conv-layer.md
```diff
@@ -199,8 +199,8 @@ conv2d = nn.Conv2D(1, kernel_size=(1, 2))
 conv2d.initialize()
 
 # The two-dimensional convolutional layer uses four-dimensional input and
-# output in the format of (example channel, height, width), where the batch
-# size (number of examples in the batch) and the number of channels are both 1
+# output in the format of (batch_size, channels, height, width), where the
```
astonzhang: Thanks. I feel it would be better if you just append a comma to "example" (before "channel")

prasanth5reddy:
@astonzhang Thanks! I did the change like you suggested

prasanth5reddy:
Did one more change in padding-and-strides.md that fixes padding example matrices' shape

PR #310, opened 2019-05-25
fix alexnet mistake according to mxnet original code

ati-ozgur:
According to original mxnet alexnet code, nn.Flatten() line is missing.

See original code:
https://mxnet.apache.org/versions/master/_modules/mxnet/gluon/model_zoo/vision/alexnet.html


astonzhang:
Thanks. This is optional because Conv will be automatically flattened before passing into Dense.

PR #317, opened 2019-05-30
refine autograd docs

larroy:


larroy:
@mli is there a way to see the book output without having to compile locally and execute all the notebooks? is very slow in my machine.

mli:
You can click the "All notebook" link on the navbar, or here is the link: https://en.d2l.ai/d2l-en.zip

larroy:
I meant from a PR.

chapter_crashcourse/autograd.md
```diff
@@ -35,7 +35,7 @@ with autograd.record():
 print(y)
 ```
 
-Since the shape of `x` is (4, 1), `y` is a scalar. Next, we can automatically find the gradient by calling the `backward` function. It should be noted that if `y` is not a scalar, MXNet will first sum the elements in `y` to get the new variable by default, and then find the gradient of the variable with respect to `x`.
+Since the shape of `x` is (4, 1), `y` is a scalar. Next, we can automatically find the gradient of all the inputs by calling the `backward` function. It should be noted that if `y` is not a scalar, MXNet will first sum the elements in `y` to get the new variable by default, and then find the analytical gradient of the variable with respect to `x` evaluated at `x` that is $\frac{dy}{dx}\bigr\rvert_{x=x_0}$.
```
astonzhang: x_0 is undefied

larroy:
When I execute d2lbook it executes the notebooks, is it possible to skip them? in CI seems you go much faster.

mli:
You can change `eval_notebook = True` to False in config.ini to disable notebook execution. I'll add "push PR build results to a staging repo" as a TODO

PR #320, opened 2019-05-31
Use a separated chapter for transposed convolution 

mli:


mli:
runs fine in local, try to merge first and then debug

PR #327, opened 2019-06-06
optimization and use-gpu

haojin2:


haojin2:
@mli Should be ready for merge.

PR #328, opened 2019-06-07
Hybridize and lang-model-dataset notebooks

haojin2:
as title.


haojin2:
@mli Should be ready for merge.

PR #338, opened 2019-06-12
Reference word-level RNN better

adursun:
Even though the figure presents a word-level RNN architecture, it was referenced as a character-level one.


mli:
the CI error is due to I modified the d2l package, but hasn't update RNN chapter yet. Will fix it shortly.

PR #342, opened 2019-06-13
Add Chapter Convolutional NN and Chapter Convolutional Modern

haojin2:


d2l/train.py
```diff
@@ -240,8 +240,8 @@ def train_ch5(net, train_iter, test_iter, batch_size, trainer, ctx,
             l.backward()
             trainer.step(batch_size)
             y = y.astype('float32')
-            train_l_sum += l.asscalar()
-            train_acc_sum += (y_hat.argmax(axis=1) == y).sum().asscalar()
+            train_l_sum += l
```
reminisce: Need to call `float` on `l` to enforce the sync for this iteration. Otherwise, the memory may be exhausted if there are too many batches per epoch.
`train_l_sum += float(l)`.

d2l/train.py
```diff
@@ -240,8 +240,8 @@ def train_ch5(net, train_iter, test_iter, batch_size, trainer, ctx,
...
+            train_acc_sum += (y_hat.argmax(axis=1) == y).sum()
```
reminisce: Same here.
`train_acc_sum += float((y_hat.argmax(axis=1) == y).sum())`

PR #343, opened 2019-06-13
End math mode earlier

adursun:


mli:
Thanks for your PR. I did a major rewrite for the first part of Ch8 on https://github.com/d2l-ai/d2l-en/pull/345 This issue is fixed in it.

PR #351, opened 2019-06-16
Update sentiment

mli:


mli:
CI error should be fixed by #352

PR #367, opened 2019-06-26
NDArray to ndarray and minor changes

goldmermaid:


chapter_crashcourse/ndarray.md
```diff
@@ -25,29 +25,29 @@ x = np.arange(12)
 x
 ```
 
-When we print `x`, we can observe the property `<NDArray 12 @cpu(0)>` listed, which indicates that `x` is a one-dimensional array of length 12 and that it resides in CPU main memory. The 0 in `@cpu(0)` has no special meaning and does not represent a specific core.
+When we print `x`, we can observe the property `<ndarray 12 @cpu(0)>` listed, which indicates that `x` is a one-dimensional ndarray of length 12 and that it resides in CPU main memory. The 0 in `@cpu(0)` has no special meaning and does not represent a specific core.
```
astonzhang: http://numpy.d2l.ai/chapter_crashcourse/ndarray.html
1. no property `<NDArray 12 @cpu(0)>`
2. explain by default storing in main memory for CPU computing
3. avoid discussing @cpu(0)

chapter_crashcourse/ndarray.md
```diff
@@ -68,14 +68,14 @@ We can create tensors with each element set to 1 works via
...
-We can also specify the value of each element in the desired NDArray by supplying a Python list containing the numerical values.
+We can also specify the value of each element in the desired ndarray by supplying a Python list containing the numerical values.
 
 ```{.python .input  n=6}
 y = np.array([[2, 1, 4, 3], [1, 2, 3, 4], [4, 3, 2, 1]])
 y
 ```
 
-In some cases, we will want to randomly sample the values of each element in the NDArray according to some known probability distribution. This is especially common when we intend to use the array as a parameter in a neural network. The following snippet creates an NDArray with a shape of (3,4). Each of its elements is randomly sampled in a normal distribution with zero mean and unit variance.
+In some cases, we will want to randomly sample the values of each element in the ndarray according to some known probability distribution. This is especially common when we intend to use the array as a parameter in a neural network. The following snippet creates an ndarray with a shape of (3,4). Each of its elements is randomly sampled in a normal distribution with zero mean and standard deviation.
```
astonzhang: one standard deviation

chapter_crashcourse/ndarray.md
```diff
@@ -1,22 +1,22 @@
...
-Here, `nd` is short for `ndarray`.
+We begin by importing `np` module and `npx` module from MXNet. Here, `np` module includes all the functions that NumPy has, while `npx` module contains all the other extended functions which empower deep learning computation.
 
 ```{.python .input  n=1}
 from mxnet import np, npx
 npx.set_np()
 ```
 
-NDArrays represent (possibly multi-dimensional) arrays of numerical values. NDArrays with one axis correspond (in math-speak) to *vectors*. NDArrays with two axes correspond to *matrices*. For arrays with more than two axes, mathematicians do not have special names---they simply call them *tensors*.
+An ndarray represents (possibly multi-dimensional) arrays of numerical values. An ndarray with one axis corresponds (in math-speak) to *vectors*. An ndarray with two axes corresponds to *matrices*. For arrays with more than two axes, mathematicians do not have special names---they simply call them *tensors*.
```
astonzhang: When using ndarray, we almost always invoke the `set_up` function: this is for compatibility of ndarray processing by other components of MXNet.

chapter_crashcourse/ndarray.md
```diff
@@ -132,23 +133,24 @@ For stylistic convenience, we can write `y.exp()`, `x.sum()`, `x.norm()`, etc. a
...
 ```
 
-Since `a` and `b` are (3x1) and (1x2) matrices respectively, their shapes do not match up if we want to add them. NDArray addresses this by 'broadcasting' the entries of both matrices into a larger (3x2) matrix as follows: for matrix `a` it replicates the columns, for matrix `b` it replicates the rows before adding up both element-wise.
+Since `a` and `b` are (3x1) and (1x2) matrices respectively, their shapes do not match up if we want to add them. We 'broadcast' the entries of both matrices into a larger (3x2) matrix as follows: for matrix `a` it replicates the columns, for matrix `b` it replicates the rows before adding up both element-wise.
 
 ```{.python .input}
 a + b
 ```
 
 ## Indexing and Slicing
```
astonzhang: At the beginning, show example of `x[1]` and highlight that its result is a scalar.

chapter_crashcourse/ndarray.md
```diff
@@ -197,9 +199,9 @@ x += y
 id(x) == before
 ```
 
-## Mutual Transformation of NDArray and NumPy
+## `mxnet.numpy.ndarray` and `numpy.ndarray`
 
-Converting MXNet NDArrays to and from NumPy is easy. The converted arrays do *not* share memory. This minor inconvenience is actually quite important: when you perform operations on the CPU or one of the GPUs, you do not want MXNet having to wait whether NumPy might want to be doing something else with the same chunk of memory. The  `array` and `asnumpy` functions do the trick.
+Converting MXNet ndarrays to and from NumPy is easy. The converted arrays do **not** share memory. This minor inconvenience is actually quite important: when you perform operations on the CPU or one of the GPUs, you do not want MXNet having to wait whether NumPy might want to be doing something else with the same chunk of memory. The  `array` and `asnumpy` functions do the trick.
```
astonzhang: Highlight that NumPy here is a scientific computing package of Python, not of MXNet.

chapter_crashcourse/ndarray.md
```diff
@@ -1,22 +1,22 @@
...
 
-Throughout this chapter, we are aiming to get you up and running with the basic functionality. Do not worry if you do not understand all of the basic math, like element-wise operations or normal distributions. In the next two chapters we will take another pass at the same material, teaching the material in the context of practical examples. On the other hand, if you want to go deeper into the mathematical content, see :numref:`chapter_math`.
+Throughout this chapter, we are aiming to get you up and running with the basic functionality. Do not worry if you do not understand all of the basic math, like element-wise operations or normal distributions. In the next two chapters we will take another pass at the same material with practical examples. On the other hand, if you want to go deeper into the mathematical content, see :numref:`chapter_math`.
 
-We begin by importing MXNet and the `ndarray` module from MXNet.
-Here, `nd` is short for `ndarray`.
+We begin by importing `np` module and `npx` module from MXNet. Here, `np` module includes all the functions that NumPy has, while `npx` module contains all the other extended functions which empower deep learning computation.
```
astonzhang: NumPy (a scientific computing package of Python)

chapter_crashcourse/ndarray.md
```diff
@@ -1,22 +1,22 @@
...
+It is impossible to get anything done if we cannot manipulate data. Generally, there are two important things we need to do with data: (i) acquire it and (ii) process it once it is inside the computer. There is no point in acquiring data if we do not even know how to store it, so let's get our hands dirty first by playing with synthetic data. We will start by introducing the N-dimensional array (ndarray), MXNet's primary tool for storing and transforming data. 
+If you have worked with NumPy's ndarray before, you will notice that the MXNet's ndarray is an extension to the former. However, the latter provides a few key advantages. First, it supports asynchronous computation on CPU, GPU, and distributed cloud architectures, whereas the former only supports CPU computation. Second, MXNet's ndarray supports for automatic differentiation. These properties make MXNet's ndarray indispensable for deep learning.
```
astonzhang: explain NumPy is a scientific computing package of Python.

PR #369, opened 2019-06-26
[numpy] Fix chapters CV and GAN

reminisce:
Depending on https://github.com/apache/incubator-mxnet/pull/15368.


reminisce:
CI is failing due to something wrong with multi-GPU training in MXNet using numpy ops. Single GPU can run through. Debugging...

PR #398, opened 2019-07-07
Fix a couple of typos

abhinav-upadhyay:
The line 

> this doesn't means that there cannot be more than one global minimum...

 should be "this doesn't mean that there cannot be more than one global minimum..."
Also,  the spelling of define was wrong on another line, it was spelled as "defing"


goldmermaid:
LGTM @astonzhang

PR #400, opened 2019-07-09
Fix textCNN multichannel initialization

leezu:
According to https://arxiv.org/pdf/1408.5882.pdf Section 3.3 both channels are initialized from the pre-trained embeddings.


astonzhang:
Thanks Leo! The Chinese version is correct: http://zh.d2l.ai/chapter_natural-language-processing/sentiment-analysis-cnn.html

PR #401, opened 2019-07-11
Fixed a couple of wording issues in the preface chapter

BigBubba:
Looked like a couple of spots where partial updates were made to sentences and parts of both were in the text.


goldmermaid:
LGTM @astonzhang

PR #419, opened 2019-08-06
Fixes to Appendices

qbaza:
1.) x^T A x can be calculated only for a square matrix A with dimension equal to length of x
2.) Hessian matrix is square n x n and not n x m


qbaza:
Fixed the image references in section 15.3, Appendix: Jupyter

PR #422, opened 2019-08-20
Update why-conv.md

dchoi77:
The equations in 6.1.2 do not express a bias term explicitly and class Conv2D implemented in 6.2.2 has a bias term. I think that it would be better if you express a bias term in 6.1.2 clearly. For example, we can add u[i,j] to the summation in each equation. By translation invariance, it's clear that u[i,j] does not depend on i or j. Thus u[i,j] is a constant.

It would be better to use b[i,j] instead of u[i,j], but the symbol b is already used as index of summation.


goldmermaid:
LGTM.

PR #424, opened 2019-08-20
Update softmax-regression-gluon.md

dchoi77:
1) What does y_linear mean?
2) $\hat{y}$ is the softmax.


goldmermaid:
Good catch. Please change the first one to:

"is the j-th element of ``yhat`` and $z_j$ is the j-th element of the input ``z`` variable, as computed by the z = 𝐖𝐱+𝐛."

goldmermaid:
LGTM

PR #425, opened 2019-08-20
Update softmax-regression-scratch.md

dchoi77:
I am not sure why nrows and ncols are needed as input parameters of class Animator, but I would like to point out something about it. Animator works well when nrows is 1 or ncols is 1, but it does not work when both nrows and ncols are greater than 1, since self.axes[0].cla() will raise an error.
To prevent it, we may add the following line right after the line if nrows * ncols == 1: self.axes = [self.axes,]:

if nrows > 1 and ncols > 1: self.axes = self.axes.flatten()

### DIFF
--- a/chapter_linear-networks/softmax-regression-scratch.md
+++ b/chapter_linear-networks/softmax-regression-scratch.md
@@ -282,6 +282,7 @@ class Animator(object):
         d2l.use_svg_display()
         self.fig, self.axes = d2l.plt.subplots(nrows, ncols, figsize=figsize)
         if nrows * ncols == 1: self.axes = [self.axes,]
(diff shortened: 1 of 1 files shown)


goldmermaid:
"plt.subplots" is calling from "matplotlib.pyplot.subplots", please refer to https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.subplots.html#matplotlib.pyplot.subplots

The reason we add "if nrows * ncols == 1: self.axes = [self.axes,]" is the above function will automatically "squeeze" if only one subplot is constructed (nrows=ncols=1), the resulting single Axes object is returned as a scalar.

PR #442, opened 2019-08-29
Update anchor.md

dchoi77:
I double checked the return values of the function MultiBoxPrior. I noticed that if [a, b, c, d] is for an anchor box returned by the function, then (c - a) * w which is the width of the anchor box  is not w * s * sqrt(r) but delta * w * s * sqrt(r), where delta is a constant and about 0.7706. The height of the anchor box is (d - b) * h which is equal to h * s / sqrt(r) as expected.
It would be nicer if you explain such a difference.


astonzhang:
Thanks. To review your changes, could you:

1) Save your changes;
2) Delete your forked yourusername/d2l-en repo, re-fork d2l-ai/d2l-en to yourusername/d2l-en
3) Commit your changes to yourusername/d2l-en, re-submit your PR.
4) Then you may close this PR so we can review your new PR

astonzhang:
Closing this PR for now. You could re-submit this PR by following the instructions above. Thanks!

PR #456, opened 2019-09-23
Minor corrections

eric-haibin-lin:
they use the [two-sample test](https://en.wikipedia.org/wiki/Two-sample_hypothesis_testing) to provide training signal to a generative model ->
they use the [two-sample test](https://en.wikipedia.org/wiki/Two-sample_hypothesis_testing) to provide training **signals** to a generative model

At the very least, it needs to fool the classifier. And if our classifier is a state of the art deep neural network. -> 
At the very least, it needs to fool the classifier. **Even** if our classifier is a state of the art deep neural network.


eric-haibin-lin:
Changed to numpy2 branch

PR #498, opened 2019-10-22
RNN proofread first two sections

goldmermaid:


astonzhang:
Can you solve the conflict due to my recent fix?

PR #547, opened 2019-11-03
AB testing first draft

goldmermaid:


goldmermaid:
First draft is done. @astonzhang 

However, it cannot pass the Jenkins without scipy package. I have to use a function in scipy:

> scipy.stats.norm.ppf 

since Hao said it is hard to implement it with the time limit.

PR #561, opened 2019-11-07
replace | with \mid

goldmermaid:


goldmermaid:
With a little bit "p(x)" to "P(x)"

PR #569, opened 2019-11-10
att9.2 fully revised

goldmermaid:


chapter_attention-mechanism/seq2seq-attention.md
```diff
@@ -3,20 +3,24 @@
...
-attention layer consists of the encoder outputs of each time step. During
-decoding, the decoder output from the previous time step is used as the query,
-the attention output is then fed into the decoder with the input to provide
-attentional context information.
+architecture for encoding and decoding at the time step $t$. Here, the memory of the
+attention layer consists of all the information that the encoder has 
+seen---the encoder outputs at each time step. 
+During the decoding, the decoder output from the previous time step $t-1$ is used as the query.
+While the output of the attention model is viewed as the context information, 
+which is then concatenated with the decode input $D_t$. Finally, we feed the concatenation into the decoder.
```
astonzhang: decode input -> decoder input

astonzhang: During the decoding, the decoder output from the previous time step $t-1$ is used as the query.
While the output of the attention model is viewed as the context information, 
which is then concatenated with the decode input $D_t$.

hard to understand

chapter_attention-mechanism/seq2seq-attention.md
```diff
@@ -88,12 +94,13 @@ out.shape, len(state), state[0].shape, len(state[1]), state[1][0].shape
 
 ## Training
 
-Again, we use the same training hyper-parameters as in
-:numref:`sec_seq2seq`. The training loss is similar to the seq2seq model, because the
+Similar to :numref:`sec_seq2seq_training`, we try a toy model by applying
+the same training hyper-parameters and the same training loss.
+As we can see from the result, since the
 sequences in the training dataset are relative short. The additional attention
```
astonzhang: As we can see from the result, since the
sequences in the training dataset are relative short.


Incomplete sentence

PR #581, opened 2019-11-12
python 3.7 for installation

goldmermaid:


chapter_appendix_tools/aws.md
```diff
@@ -231,70 +231,57 @@ First, to simplify the installation, you need to install [Miniconda](https://con
...
+conda create --name d2l -y
 ```
 
 
-![ Conda create environment D2L. ](../img/conda_create_d2l.png)
-:width:`700px`
-:label:`fig_conda_create_d2l`
 
-After create `d2l` environment, activate it and install `pip`.
+After create `d2l` environment, activate it and install `pip` and `python`. Note that currently we need to install python3.7, since python3.8 is still at its testing period and is not  compatible with some packages.
```
astonzhang: remove: Note that currently we need to install python3.7, since python3.8 is still at its testing period and is not  compatible with some packages.

PR #595, opened 2019-11-14
update

cheungdaven:


chapter_recommender-systems/autorec.md
```diff
@@ -122,7 +122,3 @@ d2l.train_recsys_rating(net, train_iter, test_iter, loss, trainer, num_epochs,
 ## References
```
astonzhang: we should remove such references, and cite more properly

PR #566, opened 2019-11-08
chore: add .gitattributes and set eof=lf

jjangga0214:
Motivated by https://github.com/d2l-ai/data/pull/2. This will prevent CRLF to be committed.

PR #664, opened 2019-12-13
chore: add .gitattributes and set eof=lf

jjangga0214:
_This was originally #566, but reposted here by an author's request._

Motivated by https://github.com/d2l-ai/data/pull/2. This will prevent CRLF to be committed.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


astonzhang:
Thanks. Since in Windows both a CR and LF are required to note the end of a line, maybe we don't need this config?

jjangga0214:
@astonzhang This will ensure files are automatically converted to LF, only when committed and copied to .git directory. So only "copied" files are affected. 
So this config does not convert CRLF files to LF from the local workspace. Such an effect, which is not a concern of .gitattribute, can be achieved by [editorconfig](https://editorconfig.org/).
When Windows users check out to branch, clone repo, pull new files, etc, their git will decide whether to copy files to the workspace from .git with LF or CRLF by their personal configuration. Though I don't remember default config for Windows, as far as I know, this is a safe configuration.

PR #665, opened 2019-12-15
fix featrue typo in chapter_modern-convolutional-neural-networks/alex…

RichyChen:
…net.md

*Description of changes:*
 I am reading the <Dive into deep learning> book, and find a typo in the "Deep Convolutional Neural Networks (AlexNet)".
    The original text is : 
      Interestingly in the lowest layers of the network, the model learned featrue extractors that resembled some traditional filters.

     I think it should be "feature extractors".

choice.

### DIFF
--- a/chapter_convolutional-neural-networks/conv-layer.md
+++ b/chapter_convolutional-neural-networks/conv-layer.md
@@ -186,7 +186,7 @@ Next, in each iteration, we will use the squared error
 to compare `Y` and the output of the convolutional layer,
 then calculate the gradient to update the weight.
 For the sake of simplicity, in this convolutional layer,
-we will ignores the bias.
+we will ignore the bias.
 
 We previously constructed the `Conv2D` class.
 However, since we used single-element assignments,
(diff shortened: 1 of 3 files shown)


astonzhang:
Thanks!

RichyChen:
OK. However, I find another typo in https://d2l.ai/chapter_convolutional-neural-networks/padding-and-strides.html 

> 
In some cases, we want to reduce the resolution drastically if say we find our original input resolution to be _unweildy_. Strides can help in these instances.
> 
I think it should be **unwieldy**.

astonzhang:
Thanks. Can you send another PR?

RichyChen:
Sorry, maybe another unproper expression on the same page  https://d2l.ai/chapter_convolutional-neural-networks/padding-and-strides.html#id1
"For any two-dimensional array X, when the kernels size is odd and the number of padding rows and columns on all sides are the same, producing **an output with _the have_ the same height and width as the input**"
It should be **an output with the same height and width as the input** .

RichyChen:
Sure, I'll send.

PR #711, opened 2020-01-27
[WIP] Formula in rec/ranking

goldmermaid:
To discuss:

- In formula 15.5.2, should it be "\propto" or "="?
- In formula 15.5.2, how to derive the last two steps? Did I miss the definition of $\lambda_\Theta$?


astonzhang:
@cheungdaven Can you take a look at it?

cheungdaven:
Hi @goldmermaid, it should be \propto not "=", thanks for pointing out.

PR #713, opened 2020-01-28
small fix to introduction chapter

karolszk:
In classification section added last letter 'z' to character set.


goldmermaid:
Hi @karolszk , thanks for your feedbacks. Here, we are not limited to characters and numbers. For example, latin characters may be considered. Please close the PR if you don't have any other question.

PR #739, opened 2020-02-18
Minor fixes in 1, 8.5, 9.6, 9.8

sanjaradylov:


goldmermaid:
Hi @sanjaradylov , thanks for your contribution! Great catches!

LGTM @astonzhang

chapter_recurrent-modern/encoder-decoder.md
```diff
@@ -21,7 +21,7 @@ class Encoder(nn.Block):
     def __init__(self, **kwargs):
         super(Encoder, self).__init__(**kwargs)
 
-    def forward(self, X):
+    def forward(self, X, *args):
```
astonzhang: Do we absolutely really need *args? If not, we can keep it simple by removing it.

sanjaradylov: According to `EncoderDecoder`, `forward` accepts encoder and decoder inputs as well as additional arguments depending on context (e.g. valid length), then passes these arguments into `Encoder`'s `forward` and `Decoder`'s `init_state`. Since we cannot instantiate an abstract class `Encoder` (instead, we create new inherited classes), the previous implementation won't cause errors. `*args` is for consistency purposes only, to describe the variable scope of encoder's forward propagation.

PR #740, opened 2020-02-20
updated order of imports

karolszk:
According to python guidelines (PEP8), standard imports should be before third-party imports


astonzhang:
Thanks. In d2l we don't have much space to separate standard imports and third-party imports. Thus, we only order alphabetically.

karolszk:
Reason understood. It worth notice that directly copying examples and executing lint software on it, for example pylint3 will cause warnings like this below:

`C0411: standard import "import random" should be placed before "import d2l" (wrong-import-order)`

so some pythonistas will complain.

PR #745, opened 2020-02-24
"as is show in in" -> "as is show in" typo

IgorDzreyev:
*Description of changes:*

choice.


IgorDzreyev:
Next to a9fe515 commit sentence is also need correction - "Do you care
about the average performance only, or do you also the range of variation of the scores."

PR #749, opened 2020-02-28
small fix to text unnecessary italic

karolszk:
Removed unnecessary italic style of conjunction 'and'


astonzhang:
italic here is to emphasize

karolszk:
ok.

PR #761, opened 2020-03-04
modified class Timer  to be more pythonable 3

karolszk:
Hello,

all classes in Python3 inherits from object() class by design. Thus I removed unnecessary inheritance.
Ref doc: https://docs.python.org/3/library/functions.html?highlight=object%20class#object


astonzhang:
In this case, we'll force users to install python 3, not 2

karolszk:
Yes, support for python2 officially stopped in Jan 2020, https://www.python.org/dev/peps/pep-0373/ so forcing 3 is good decision.

PR #769, opened 2020-03-07
handling path building in xplatform manner + some minor issues and typos

IgorDzreyev:
*Description of changes:*

choice.

### DIFF
--- a/chapter_computer-vision/kaggle-gluon-cifar10.md
+++ b/chapter_computer-vision/kaggle-gluon-cifar10.md
@@ -105,16 +105,16 @@ def reorg_train_valid(data_dir, labels, valid_ratio):
     label_count = {}
     for train_file in os.listdir(data_dir + 'train'):
         label = labels[train_file.split('.')[0]]
-        fname = data_dir + 'train/' + train_file
+        fname = os.path.join(data_dir, 'train', train_file)
         # Copy to train_valid_test/train_valid with a subfolder per class
-        copyfile(fname, data_dir + 'train_valid_test/train_valid/' + label)
+        copyfile(fname, os.path.join(data_dir, 'train_valid_test','train_valid', label))
         if label not in label_count or label_count[label] < n_valid_per_label:
             # Copy to train_valid_test/valid
-            copyfile(fname, data_dir + 'train_valid_test/valid/' + label)
+            copyfile(fname, os.path.join(data_dir, 'train_valid_test', 'valid', label))
             label_count[label] = label_count.get(label, 0) + 1
         else:
             # Copy to train_valid_test/train
-            copyfile(fname, data_dir+'train_valid_test/train/' + label)
+            copyfile(fname, os.path.join(data_dir, 'train_valid_test', 'train', label))
     return n_valid_per_label
 ```
 
@@ -124,8 +124,8 @@ The `reorg_test` function below is used to organize the testing set to facilitat
 # Saved in the d2l package for later use    
 def reorg_test(data_dir):
     for test_file in os.listdir(data_dir + 'test'):
-        copyfile(data_dir + 'test/' + test_file, 
-                 data_dir + 'train_valid_test/test/unknown/')
+        copyfile(os.path.join(data_dir, 'test', test_file),
+                 os.path.join(data_dir, 'train_valid_test', 'test', 'unknown'))
 ```
 
 Finally, we use a function to call the previously defined `read_csv_labels`, `reorg_train_valid`, and `reorg_test` functions.
--- a/chapter_computer-vision/kaggle-gluon-cifar10.md
+++ b/chapter_computer-vision/kaggle-gluon-cifar10.md
@@ -81,7 +81,7 @@ def read_csv_labels(fname):
     tokens = [l.rstrip().split(',') for l in lines]
     return dict(((name, label) for name, label in tokens))
 
-labels = read_csv_labels(data_dir + 'trainLabels.csv')
+labels = read_csv_labels(os.path.join(data_dir, 'trainLabels.csv'))
 print('# training examples:', len(labels))
 print('# classes:', len(set(labels.values())))
 ```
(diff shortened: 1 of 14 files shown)


astonzhang:
Can you resolve the conflicts by removing (object)?

IgorDzreyev:
> Can you resolve the conflicts by removing (object)?
Looks like it is done...

astonzhang:
Thanks!
A minor suggestion is that, you may wish to install an nbextension called ruler to ensure that each line in the code cell does not exceed 78 chars:
https://d2l.ai/chapter_appendix-tools-for-deep-learning/jupyter.html#timing

astonzhang:
Regarding your change:

natural-language-inference-and-dataset.md
        s = re.sub('\(', '', s) 
        s = re.sub('\)', '', s)
->
        s = re.sub('\\(', '', s) 
        s = re.sub('\\)', '', s)

        s = re.sub('\s{2,}', ' ', s)
->
        s = re.sub('\\s{2,}', ' ', s)

For input
s = '( ( ( A person ) ( on ( a horse ) ) ) ( ( jumps ( over ( a ( broken ( down airplane ) ) ) ) ) . ) )'

both correctly return
'A person on a horse jumps over a broken down airplane .'

Is there any reason why \\ is preferred over \?

IgorDzreyev:
Well, I've noticed this with help of PyCharm.

Here is what Official Python Docs says (https://docs.python.org/3.8/library/re.html):

...
Regular expressions use the backslash character ('\') to indicate special forms or to allow special characters to be used without invoking their special meaning. This collides with Python’s usage of the same character for the same purpose in string literals; for example, to match a literal backslash, one might have to write '\\\\' as the pattern string, because the regular expression must be \\, and each backslash must be expressed as \\ inside a regular Python string literal. Also, please note that any invalid escape sequences in Python’s usage of the backslash in string literals now generate a DeprecationWarning and in the future this will become a SyntaxError. This behaviour will happen even if it is a valid escape sequence for a regular expression.
...

astonzhang:
Thanks. Besides, after your changes in `download_extract`, we also need to make some more changes:
https://github.com/d2l-ai/d2l-en/commit/50ad5dea5e4edb95193cd55ee8928d8c841da294
https://github.com/d2l-ai/d2l-en/commit/93de46ee9bedd46bdc4dc68d9f6845342a69835c

to pass the entire book. This is due to the fact that d2l.download_extract will be invoked by many other notebooks, thus changes of anything inside d2l.py needs to be evaluated throughout the entire book. Unfortunately, CI only tests your modified files thus misses to report errors in other affected notebooks.

To test changes that potentially affect the entire book, you may 

1. `pip install d2lbook`
1. `cd d2l-en && python setup.py develop  # So any changes in d2l/d2l.py will take effect`
1. `d2lbook build lib  # Copy changes in d2l functions from md source to d2l/d2l.py`
1.   In config.ini, set `force_eval_entire_book = True`
1. `d2lbook build eval  # Test the entire book`
1. If the entire book passes, `git checkout -- config.ini`

IgorDzreyev:
Yes, I've tried to build it all locally, but there is an issue with multiple GPU chapter - it requires 2 GPUs...

IgorDzreyev:
Regarding d2l.py - Isn't it build from chapters' code?

IgorDzreyev:
'...Unfortunately, CI only tests your modified files thus misses to report errors in other affected notebooks...' - after I've made first commit and couple after - CI failed in different places I didn't touch, so it looks like it checks all the code after all - those changes with "import os" were for that

IgorDzreyev:
"d2lbook build eval" just converts .md-s to .ipynb-s... it there a command in d2lbook to actually execute all notebooks?

astonzhang:
> Regarding d2l.py - Isn't it build from chapters' code?

yes, but you need to explicitly `d2lbook build lib` to copy from source to lib




> "d2lbook build eval" just converts .md-s to .ipynb-s... it there a command in d2lbook to actually execute all notebooks?

https://github.com/d2l-ai/d2l-en/blob/master/config.ini#L35
should be True

astonzhang:
regarding those 2 GPU requirement, you may exclude them in your eval
https://github.com/d2l-ai/d2l-en/blob/master/config.ini#L32

IgorDzreyev:
Cool, thanks for clarifying!
By any chance - is there "-j option" for parallel evaluation? )

astonzhang:
that's a nice idea! I think maybe I can support that feature by using Python multiprocessing. Do you have any suggestions?

IgorDzreyev:
Well, I didn't have a change to look into d2lbook code thoroughly...

IgorDzreyev:
https://github.com/d2l-ai/d2l-book/blob/a75585d41f7fb6e39866783fd65bfa6b7116e4c2/d2lbook/build.py#L427 ?

IgorDzreyev:
Hmm... Some of notepads loads GPU at top. So it can be an issue when running them simultaneously... They will fail with 'Out of (GPU) RAM'...

On the other hand - there is not much of them, so they may be treated separately - serialized, at the end of evaluation, for example.

But the question - will it be worth of maintaining all this stuff...

IgorDzreyev:
Hi!
Ran everything except chapter_computational-performance/multiple-gpus-gluon.md and chapter_computational-performance/multiple-gpus.md

Has only one failure - chapter_natural-language-processing-applications/natural-language-inference-bert.md. Figuring out why.

IgorDzreyev:
Something about:

UserWarning: Gradient of Parameter `dense76_bias` on context gpu(0) has not been updated by backward since last `step`. This could mean a bug in your model that made it only use a subset of the Parameters (Blocks) for this iteration. If you are intentionally only using a subset, call step with ignore_stale_grad=True to suppress this warning and skip updating of Parameters with stale gradient

IgorDzreyev:
I rolled way back before my commits and this error is there...

PR #809, opened 2020-04-02
Compatibility changes to master branch on 20200312

haojin2:
As title.
@astonzhang


astonzhang:
notes: when np.array([1]) is integer, describe it in ndarray.md

PR #837, opened 2020-04-28
birnn "forward pass" to "forward propagation"

goldmermaid:


astonzhang:
I'll make some further edits

PR #879, opened 2020-05-20
pytorch: add softmax scratch

AnirudhDagar:


chapter_linear-networks/softmax-regression-scratch.md
```diff
@@ -9,17 +9,33 @@ As with linear regression, after doing things by hand
 we will breeze through an implementation in Gluon for comparison.
 To begin, let us import the familiar packages.
 
-```{.python .input  n=1}
+```{.python .input}
+import sys
+sys.path.insert(0, "../")
```
mli, line 14: We don’t need it as we assume users have already install d2l

PR #880, opened 2020-05-20
Fix grammar in rnn.md

cuongvng:
*Description of changes:*
Fix a small grammar error.

choice.

### DIFF
--- a/chapter_recurrent-neural-networks/rnn.md
+++ b/chapter_recurrent-neural-networks/rnn.md
@@ -12,7 +12,7 @@ $$h_t = f(x_{t}, h_{t-1}).$$
 
 For a sufficiently powerful function $f$, the latent variable model is not an approximation. After all, $h_t$ could simply store all the data it observed so far. We discussed this in :numref:`sec_sequence`. But it could potentially makes both computation and storage expensive.
 
-Note that we also use $h$ to denote the number of hidden units of a hidden layer. Hidden layers and hidden states refer to two very different concepts. Hidden layers are, as explained, layers that are hidden from view on the path from input to output. Hidden states are technically speaking *inputs* to whatever we do at a given step. Instead, they can only be computed by looking at data at previous iterations. In this sense they have much in common with latent variable models in statistics, such as clustering or topic models where the clusters affect the output but cannot be directly observed.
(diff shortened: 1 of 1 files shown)


chapter_recurrent-neural-networks/rnn.md
```diff
@@ -12,7 +12,7 @@ $$h_t = f(x_{t}, h_{t-1}).$$
...
+Note that we also use $h$ to denote the number of hidden units of a hidden layer. Hidden layers and hidden states refer to two very different concepts. Hidden layers are, as explained, layers that are hidden from view on the path from input to output. Hidden states are technically speaking *inputs* to whatever we do at a given step. Instead, they can only be computed by looking at data at previous iterations. In this sense they have much in common with latent variable models in statistics, such as clustering or topic models where the clusters affect the output but cannot be directly observed.
```
cuongvng, line 15:
```suggestion
Note that we also use $h$ to denote the array of hidden units of a hidden layer. Hidden layers and hidden states refer to two very different concepts. Hidden layers are, as explained, layers that are hidden from view on the path from input to output. Hidden states are technically speaking *inputs* to whatever we do at a given step. Instead, they can only be computed by looking at data at previous iterations. In this sense they have much in common with latent variable models in statistics, such as clustering or topic models where the clusters affect the output but cannot be directly observed.
```

chapter_recurrent-neural-networks/rnn.md
```diff
@@ -12,7 +12,7 @@ $$h_t = f(x_{t}, h_{t-1}).$$
...
+Note that we also use $h$ to denote the array of hidden units of a hidden layer. Hidden layers and hidden states refer to two very different concepts. Hidden layers are, as explained, layers that are hidden from view on the path from input to output. Hidden states are technically speaking *inputs* to whatever we do at a given step. Instead, they can only be computed by looking at data at previous iterations. In this sense they have much in common with latent variable models in statistics, such as clustering or topic models where the clusters affect the output but cannot be directly observed.
```
cuongvng, line 15:
```suggestion
Note that we also use $h$ to denote the number of hidden units of a hidden layer. Hidden layers and hidden states refer to two very different concepts. Hidden layers are, as explained, layers that are hidden from view on the path from input to output. Hidden states are technically speaking *inputs* to whatever we do at a given step. Instead, they can only be computed by looking at data at previous iterations. In this sense they have much in common with latent variable models in statistics, such as clustering or topic models where the clusters affect the output but cannot be directly observed.
```

PR #922, opened 2020-06-01
pytorch: add sequence model

AnirudhDagar:
*Description of changes:*

choice.

### DIFF
--- a/chapter_recurrent-neural-networks/sequence.md
+++ b/chapter_recurrent-neural-networks/sequence.md
@@ -145,10 +145,9 @@ def init_weights(m):
 
 # Vanilla MLP architecture
 def get_net():
-    net = nn.Sequential()
-    net.add_module('Linear_1', nn.Linear(4, 10))
-    net.add_module('ReLU', nn.ReLU())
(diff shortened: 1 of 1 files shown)


chapter_recurrent-neural-networks/sequence.md
```diff
@@ -111,6 +124,38 @@ def get_net():
...
+                           batch_size, is_train=False)
+
+# Function for initializing the weights of net
+def init_weights(m):
+    if type(m) == nn.Linear:
+        torch.nn.init.xavier_uniform_(m.weight)
+
+# Vanilla MLP architecture
+def get_net():
+    net = nn.Sequential()
```
mli, line 148: we didn't introduce `add_module` before, we can use `nn.Sequential(nn.xxx)` as much as possible

AnirudhDagar: Yes, I can go do that. I thought `add_module` is a bit more clean implementation wise. But I'll send in changes if we want to stick with `nn.Sequential(nn.xx , nn.xx , ... , nn.xx )`

PR #945, opened 2020-06-08
Minor Fix

goldmermaid:
Add ctx=x.ctx, or the training will not use a GPU.

choice.


astonzhang:
I understand that this will enable GPU running. However, this section is meant to be run on CPU so would this be unnecessarily complicated?

PR #946, opened 2020-06-08
Undefined name: import warnings for line 372

cclauss:
`warnings.warn()` is called on line 372 but `warnings` is neither imported nor defined so NameError will probably be raised instead of a warning.

*Description of changes:*

choice.

### DIFF
--- a/d2l/torch.py
+++ b/d2l/torch.py
@@ -24,6 +24,7 @@ import shutil
 import sys
 import tarfile
 import time
+import warnings
 import zipfile
 import requests
 


AnirudhDagar:
@mli please merge this. I missed this import while adding the code. Also, Jenkins did not fail because of the if condition defined here.
Thanks, @cclauss for the PR.

mli:
@AnirudhDagar please add the import the preface.md, adding to torch.py can be easily rewritten

AnirudhDagar:
Fixed in #950

PR #1007, opened 2020-06-14
pytorch: 9.5 and 9.6

ypandya:
*Description of changes:*

choice.

### DIFF
--- a/chapter_recurrent-modern/encoder-decoder.md
+++ b/chapter_recurrent-modern/encoder-decoder.md
@@ -12,11 +12,25 @@ In this section, we will show an interface to implement this encoder-decoder arc
 
 The encoder is a normal neural network that takes inputs, e.g., a source sentence, to return outputs.
 
-```{.python .input  n=2}
+```{.python .input  n=4}
 from mxnet.gluon import nn
 
 #@save
 class Encoder(nn.Block):
+    """The base encoder interface for the encoder-decoder architecture."""
+    def __init__(self, **kwargs):
+        super(Encoder, self).__init__(**kwargs)
+
+    def forward(self, X, *args):
+        raise NotImplementedError
+```
+
+```{.python .input  n=1}
+#@tab pytorch
+from torch import nn
+
+#@save
+class Encoder(nn.Module):
     """The base encoder interface for the encoder-decoder architecture."""
     def __init__(self, **kwargs):
         super(Encoder, self).__init__(**kwargs)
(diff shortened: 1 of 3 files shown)


chapter_recurrent-modern/encoder-decoder.md
```diff
@@ -25,6 +25,20 @@ class Encoder(nn.Block):
         raise NotImplementedError
 ```
 
+```{.python .input  n=1}
```
AnirudhDagar, line 28: Looks like the tabs were not cleared out. Can you restart and clear the outputs of the notebook and then run `d2lbook activate all`.

This should look like -> ```{.python .input}

chapter_recurrent-modern/machine-translation-and-dataset.md
```diff
@@ -118,13 +131,27 @@ def build_array(lines, vocab, num_steps, is_source):
...
+```{.python .input}
+#@tab pytorch
+#@save
+def build_array(lines, vocab, num_steps, is_source):
+    lines = [vocab[l] for l in lines]
+    if not is_source:
+        lines = [[vocab['<bos>']] + l + [vocab['<eos>']] for l in lines]
+    array = torch.tensor([truncate_pad(
+        l, num_steps, vocab['<pad>']) for l in lines])
+    valid_len = (array != vocab['<pad>']).sum(axis=1)
```
AnirudhDagar, line 143:
```suggestion
    valid_len = (array != vocab['<pad>']).sum(dim=1)
```
Let's use `dim` instead of `axis` just like the PyTorch documentation which uses `dim` almost everywhere.

d2l/torch.py
```diff
@@ -670,6 +670,114 @@ def load_data_time_machine(batch_size, num_steps, use_random_iter=False,
...
+
+
+# Defined in file: ./chapter_recurrent-modern/machine-translation-and-dataset.md
+def build_array(lines, vocab, num_steps, is_source):
+    lines = [vocab[l] for l in lines]
+    if not is_source:
+        lines = [[vocab['<bos>']] + l + [vocab['<eos>']] for l in lines]
+    array = torch.tensor([truncate_pad(
+        l, num_steps, vocab['<pad>']) for l in lines])
+    valid_len = (array != vocab['<pad>']).sum(axis=1)
```
AnirudhDagar, line 723: `dim`

chapter_recurrent-modern/encoder-decoder.md
AnirudhDagar: @ypandya please address these as well.

ypandya: I did follow these instructions, there was no change here even after clearing and activating. I checked that there are many such lines in most of the existing notebooks for the mxnet portions. These do not get updated by clearing the output. This is not a part of the output but a part of the input of the cell. Due to this many code cells have some number on them when we open even without any output.

AnirudhDagar: Yeah, it is definitely only the cell number so that is cool. But I saw that these change to nil when we restart and clear outputs. Anyways will merge this as Mu has already approved the PR.

AnirudhDagar:
Thanks for making the changes @ypandya :)

PR #1009, opened 2020-06-14
[TF] Fix accuracy issues in a couple of notebooks

terrytangyuan:
*Description of changes:*

This PR should fix the accuracy issues in the following sections:
* Softmax regression concise implementation
* Dropout
* MLP concise

choice.

### DIFF
--- a/chapter_linear-networks/softmax-regression-scratch.md
+++ b/chapter_linear-networks/softmax-regression-scratch.md
@@ -432,7 +432,7 @@ def train_epoch_ch3(net, train_iter, loss, updater):  #@save
 
 ```{.python .input}
 #@tab tensorflow
-def train_epoch_ch3(net, train_iter, loss, updater):  #@save
+def train_epoch_ch3(net, train_iter, loss, updater, params=None):  #@save
     metric = Accumulator(3)  # train_loss_sum, train_acc_sum, num_examples
     for X, y in train_iter:
         # Compute gradients and update parameters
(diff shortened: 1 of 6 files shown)


chapter_linear-networks/softmax-regression-scratch.md
```diff
@@ -432,7 +432,7 @@ def train_epoch_ch3(net, train_iter, loss, updater):  #@save
 
 ```{.python .input}
 #@tab tensorflow
-def train_epoch_ch3(net, train_iter, loss, updater):  #@save
+def train_epoch_ch3(net, train_iter, loss, updater, params=None):  #@save
```
mli, line 435: why we need `params`? can we remove it to make the api consistent.

terrytangyuan: In dropout section, net is a custom function which doesn’t have trainable_variables and its optimizer/updater is from tf.keras instead of using the Updater class you created in previous section. We need to pass params since we cannot attach it to either the optimizer or the custom model function.

mli: Can you use the `Updater` class instead?

terrytangyuan: Okay I switched to use `Updater` here. I originally thought that since the PyTorch implementation here uses `torch.optim.SGD` so it might be better to support `tf.keras.Optimizer` in `train_ch3()`.

mli:
Thanks

PR #1011, opened 2020-06-15
Minor Typo

goldmermaid:
The functionality is described for "validation set", but it was referred as "test set" before.

*Description of changes:*

choice.


astonzhang:
I think we are indeed referring to test data here.

PR #1013, opened 2020-06-15
Fix attention query of seq2seq.

lkhphuc:
*Description of changes:* Wrong use of `output` instead of `hidden state` of decoder's RNN. The code and the figure both say `hidden state`.

choice.

### DIFF
--- a/chapter_attention-mechanisms/seq2seq-attention.md
+++ b/chapter_attention-mechanisms/seq2seq-attention.md
@@ -38,7 +38,7 @@ Since the encoder of seq2seq with attention mechanisms is the same as `Seq2SeqEn
 
 - **the encoder valid length**: so the attention layer will not consider the padding tokens within the encoder outputs.
 
-At each timestep of the decoding, we use the output of the decoder's last RNN layer as the query for the attention layer. The attention model's output is then concatenated with the input embedding vector to feed into the RNN layer. Although the RNN layer hidden state also contains history information from decoder, the attention output explicitly selects the encoder outputs based on `enc_valid_len`, so that the attention output suspends other irrelevant information.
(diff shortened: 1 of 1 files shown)


mli:
Job d2l-en/PR-1013/4 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1013/

goldmermaid:
Hi @lkhphuc , thanks for your contribution!

The text is right, which aligns with the plot as well. We feed the hidden state of the encoder’s final timestep as the initial decoder’s hidden state; at the same time, we feed the encoder outputs as identical keys and values of the attention layer. 

However, the code was wrong, would you like to try correct it? Let me know if you feel it is challenging and I can help with it! :)

lkhphuc:
Sorry @goldmermaid, I'm not sure I follow.

According to this figure, the "query" for the attention layer is the decoder's hidden state of the previous time step. The text currently said "the output of the decoder's last RNN layer ..." which I think is wrong so I change it to "the hidden state of the decoder's last RNN...".
Screenshot 2020-06-28 at 00 20 36

I haven't work with seq2seq + Attn in detail before so I try to check with some other articles, but each one is different.
This [tutorial](https://pytorch.org/tutorials/intermediate/seq2seq_translation_tutorial.html) uses the decoder's input as query and previous hidden state as key.
This [blog](https://jalammar.github.io/visualizing-neural-machine-translation-mechanics-of-seq2seq-models-with-attention/) apply the RNN first, then use the hidden state in _this_ timestep to compute the output.

goldmermaid:
Hi @lkhphuc , no worries I will take care of it. Thanks for pointing out the typos!

mli:
Job d2l-en/PR-1013/6 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1013/

PR #1030, opened 2020-06-18
[TF] Switch to use tf.io functions for saving/loading single tensors

terrytangyuan:
*Description of changes:*

choice.

### DIFF
--- a/chapter_deep-learning-computation/read-write.md
+++ b/chapter_deep-learning-computation/read-write.md
@@ -47,7 +47,7 @@ import tensorflow as tf
 import numpy as np
 
 x = tf.constant(range(4))
-np.save("x-file.npy", x)
+tf.io.write_file("x-file", tf.io.serialize_tensor(x))
 ```
 
 We can now read this data from the stored file back into memory.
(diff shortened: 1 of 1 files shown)


mli:
Job d2l-en/PR-1030/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1030/

chapter_deep-learning-computation/read-write.md
```diff
@@ -65,7 +65,11 @@ x2
 
 ```{.python .input}
 #@tab tensorflow
-x2 = np.load("x-file.npy", allow_pickle=True)
+# Note that the out_type here is the type of the serialized
+# tensor that we wrote to a file previously. It must match the
+# type of the serialized tensor and no implicit conversion
+# will take place. 
+x2 = tf.io.parse_tensor(tf.io.read_file("x-file"), out_type=tf.int32)
```
mli, line 72: how about save a list? we still use np right now. also this APIs are not quite clean. maybe just the preview np.load/np.save approach

PR #1043, opened 2020-06-19
tf pytorch add missing numpy import

AnirudhDagar:
*Description of changes:*
numpy import is needed in pytorch and tensorflow d2l libraries as well specifically for the `Timer` class which is used later in the book. It was probably removed in [this commit](https://github.com/d2l-ai/d2l-en/commit/91072b379ebb9fdea1faf64339334f07a6a2738e#diff-7199ea3162d85ba4377898f52160727b) because mxnet uses its own version of numpy array but other libraries don't have that. Thus it is required.

choice.

### DIFF
--- a/chapter_preface/index.md
+++ b/chapter_preface/index.md
@@ -342,10 +342,10 @@ from mxnet import autograd, context, gluon, image, init, np, npx
 from mxnet.gluon import nn, rnn
 ```
 
-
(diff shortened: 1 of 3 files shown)


mli:
Job d2l-en/PR-1043/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1043/

PR #1044, opened 2020-06-19
Remove tf.cast & change torch.Tensor to torch.tensor  

AnirudhDagar:
*Description of changes:*
This PR removes the use of tf.cast in the function corr2d.
Also changes the `torch.Tensor` to `torch.tensor` which is standard across the book.
Adds explicitly the dtypes in tf and torch to make more interpretable and to remove `tf.cast` extra step in `corr2d`.

This also fixes Resnet class in tensorflow. 

choice.

### DIFF
--- a/chapter_convolutional-neural-networks/channels.md
+++ b/chapter_convolutional-neural-networks/channels.md
@@ -105,21 +105,47 @@ corr2d_multi_in(X, K)
 ```{.python .input}
 #@tab pytorch
 X = torch.tensor([[[0, 1, 2], [3, 4, 5], [6, 7, 8]],
-                  [[1, 2, 3], [4, 5, 6], [7, 8, 9]]])
+                  [[1, 2, 3], [4, 5, 6], [7, 8, 9]]], dtype=torch.float32)
 K = torch.tensor([[[0, 1], [2, 3]],
-                  [[1, 2], [3, 4]]])
+                  [[1, 2], [3, 4]]], dtype=torch.float32)
 
 corr2d_multi_in(X, K)
+```
+
+```{.json .output n=12}
+[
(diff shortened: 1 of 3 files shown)


mli:
Job d2l-en/PR-1044/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1044/

AnirudhDagar:
@mli @terrytangyuan Please review.

AnirudhDagar:
@mli @terrytangyuan This is ready for review now.

chapter_convolutional-neural-networks/channels.md
```diff
@@ -105,19 +105,19 @@ corr2d_multi_in(X, K)
 ```{.python .input}
 #@tab pytorch
 X = torch.tensor([[[0, 1, 2], [3, 4, 5], [6, 7, 8]],
-                  [[1, 2, 3], [4, 5, 6], [7, 8, 9]]])
+                  [[1, 2, 3], [4, 5, 6], [7, 8, 9]]], dtype=torch.float32)
```
mli, line 108: why not just use `torch.tensor([[[0.0, 1,...`, namely add `.0` for the first element?

AnirudhDagar: Yes. We can do that and the tensor would automatically infer the dtype. But it is always better to explicitly give dtype as an argument, which ensures that the code is more interpretable to the reader.

Also, a reader may wonder why we change only the first element to 1.0 or why we do not do such a thing in mxnet.
Thus I'd suggest explicitly mentioning dtypes wherever possible, but we can surely skip this if the input data is clearly float or int etc.

PR #1045, opened 2020-06-19
Avoid confusion between test set and validation set

goldmermaid:
From the discussion here: https://github.com/d2l-ai/d2l-en/pull/1011/files: if it was corrected, then we need to rearrange the paragraph sequences. To be concrete, we need to introduce the idea of validation set after stating the test set's functions. Or readers may feel confused and raise the issue: https://discuss.d2l.ai/t/model-selection-underfitting-and-overfitting/97


mli:
Job d2l-en/PR-1045/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1045/

PR #1053, opened 2020-06-22
[TF] Remove unnecessary softmax activation in last layer

terrytangyuan:
*Description of changes:*

By submitting this pull request, I confirm that you can use, modify,
copy, and redistribute this contribution, under the terms of your
choice.

PR #1054, opened 2020-06-22
[TF] Remove unnecessary activation layers

terrytangyuan:
Signed-off-by: terrytangyuan <terrytangyuan@gmail.com>

*Description of changes:* This is similar to previous PR https://github.com/d2l-ai/d2l-en/pull/1053 but I missed a few sections. Let's see if the CI passes.

choice.

### DIFF
--- a/chapter_convolutional-neural-networks/lenet.md
+++ b/chapter_convolutional-neural-networks/lenet.md
@@ -132,6 +132,7 @@ net = torch.nn.Sequential(
 #@tab tensorflow
 from d2l import tensorflow as d2l
 import tensorflow as tf
+from tensorflow.distribute import MirroredStrategy, OneDeviceStrategy
 
 def net():
     return tf.keras.models.Sequential([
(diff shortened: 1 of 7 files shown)


chapter_convolutional-neural-networks/lenet.md
```diff
@@ -380,11 +380,21 @@ class TrainCallback(tf.keras.callbacks.Callback):  #@save
 
 #@save
 def train_ch6(net_fn, train_iter, test_iter, num_epochs, lr,
-              device=d2l.try_gpu()):
+              device=d2l.try_gpu(), mirrored=True):
```
terrytangyuan, line 383: @mli I noticed that you removed this argument in https://github.com/d2l-ai/d2l-en/commit/3b2034c41efb226cae141796cdd1e23c7888ff2d. However, this should fail the batch norm scratch implementation as it's required in that notebook (see my comment [here](https://github.com/d2l-ai/d2l-en/pull/1054/files#diff-cf23455381e755a4d0f02d266b186b6eR533-R536)) and the CI somehow fails to catch that (perhaps d2l/tensorflow.py wasn't updated). I am adding this back for now.

mli, line 383: I feel it complicates the implementation, that's why I removed it before. Is there a simple way for TF to run on a single GPU? or we can remove the batch norm from scratch training.

terrytangyuan: I just realized that there's a [OneDeviceStrategy](https://www.tensorflow.org/api_docs/python/tf/distribute/OneDeviceStrategy) we can use. I'll give it a try.

mli: Thanks

chapter_convolutional-neural-networks/lenet.md
```diff
@@ -380,28 +381,20 @@ class TrainCallback(tf.keras.callbacks.Callback):  #@save
...
-    # machine using `tf.distribute.MirroredStrategy`.
-    if mirrored:
-        strategy = tf.distribute.MirroredStrategy(devices=[device_name])
-        with strategy.scope():
-            optimizer = tf.keras.optimizers.SGD(learning_rate=lr)
-            loss = tf.keras.losses.SparseCategoricalCrossentropy(from_logits=True)
-            net = net_fn()
-            net.compile(optimizer=optimizer, loss=loss, metrics=['accuracy'])
-    else:
+    strategy = MirroredStrategy(devices=device_name) if len(
```
mli, line 387: we can just use `OneDeviceStrategy` as we will not use multi-gpus for train_ch6. (we haven't teach it yet)

d2l/tensorflow.py
```diff
@@ -484,6 +484,7 @@ def train_ch6(net_fn, train_iter, test_iter, num_epochs, lr,
     callback = TrainCallback(net, train_iter, test_iter, num_epochs,
                              device_name)
     net.fit(train_iter, epochs=num_epochs, verbose=0, callbacks=[callback])
+    return net
```
mli, line 487: a question about `return net`. when it's needed? I found removing it works fine

terrytangyuan: There's one example here https://github.com/d2l-ai/d2l-en/pull/1054/files#diff-cf23455381e755a4d0f02d266b186b6eR552 requires accessing the trained layers/params in the model object. Since we the model is created inside the function, we will have to return it here so we can use it later.

mli: got it. thanks. the issue it causes is that jupyter will print the returned network, which is not necessary in most cases

mli:
Job d2l-en/PR-1054/7 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1054/

terrytangyuan:
@mli I’ve verified that all changed notebooks look correct. Please double check when you get a chance as well.

PR #1075, opened 2020-06-27
pytorch: fix param init

AnirudhDagar:
*Description of changes:*
This PR addresses the weight & bias initialization in section 5.2.2.1
The aim is to initialize all the layers. Hence use `m` instead of `net[0]` inside the two init functions.

choice.


AnirudhDagar:
@astonzhang Do you know why this may be failing in the pdf build ?

AnirudhDagar:
Triggered CI again.

PR #1098, opened 2020-07-02
Fix a math notation

cuongvng:
*Description of changes:*

According to the math formula on L15, I think the notation on L18 should be `g_{i,t}`, not `g_{ii}`

image

choice.


mli:
Job d2l-en/PR-1098/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1098/

goldmermaid:
Hi @cuongvng , eagle eyes! Thanks for pointing out the typos. 

However, that is not 100% what it means. I just revised it in PR https://github.com/d2l-ai/d2l-en/pull/1104. Could you review and leave a comment if it is LGTM ("Looks good to me")?

PR #1118, opened 2020-07-05
use 0 dim tensor

liweiwp:
*Description of changes:*

choice.

### DIFF
--- a/chapter_preliminaries/autograd.md
+++ b/chapter_preliminaries/autograd.md
@@ -324,9 +324,9 @@ def f(a):
 #@tab pytorch
 def f(a):
     b = a * 2
-    while b.norm().item() < 1000:
+    while b.norm() < 1000:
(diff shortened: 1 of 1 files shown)


mli:
Job d2l-en/PR-1118/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1118/

liweiwp:
> Could you explain why the changes are necessary or better than the existing code? IMO explicit is better than implicit.

@terrytangyuan @AnirudhDagar 
For the dim/size,  in MXNet version, the variable a is a scalar, but the Pytorch version is a one element vector, and the TensorFlow version is a one element matrix, which confused beginners like me, so I think the three version should consistent
For the dtype,  I think is not need for this demo, for compact code make the us focus on the key points, also the three version should consistent

PR #1146, opened 2020-07-10
polish language dataset

mli:
*Description of changes:*

choice.

### DIFF
--- a/chapter_preliminaries/ndarray.md
+++ b/chapter_preliminaries/ndarray.md
@@ -688,6 +688,7 @@ size = lambda a: a.size
 reshape = lambda a, *args: a.reshape(*args)
 ones = np.ones
 zeros = np.zeros
+tensor = np.array
 ```
 
 ```{.python .input}
@@ -698,6 +699,7 @@ size = lambda a: a.numel()
 reshape = lambda a, *args: a.reshape(*args)
 ones = torch.ones
 zeros = torch.zeros
+tensor = torch.Tensor
 ```
 
 ```{.python .input}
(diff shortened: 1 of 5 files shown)


mli:
Job d2l-en/PR-1146/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1146/

chapter_preliminaries/ndarray.md
```diff
@@ -698,6 +699,7 @@ size = lambda a: a.numel()
 reshape = lambda a, *args: a.reshape(*args)
 ones = torch.ones
 zeros = torch.zeros
+tensor = torch.Tensor
```
AnirudhDagar, line 702: @mli Not sure if this is a very good idea. Defining a tensor is always a `FloatTensor`. We won't be able to use `dtype` for defining the type explicitly with such implementation. I suggest keeping the tensors different for all the 3 frameworks. Besides this may pose problems for readers to understand.

chapter_recurrent-neural-networks/language-models-and-dataset.md
```diff
@@ -225,7 +200,7 @@ def seq_data_iter_random(corpus, batch_size, num_steps):
         batch_indices = example_indices[i:(i+batch_size)]
         X = [data(j) for j in batch_indices]
         Y = [data(j + 1) for j in batch_indices]
-        yield torch.Tensor(X), torch.Tensor(Y)
+        yield d2l.tensor(X), d2l.tensor(Y)
```
AnirudhDagar, line 203: @mli Same here. Should we really use `d2l.tensor()` instead of just using `torch.tensor()` which is more flexible.

d2l/torch.py
```diff
@@ -38,6 +38,7 @@
 reshape = lambda a, *args: a.reshape(*args)
 ones = torch.ones
 zeros = torch.zeros
+tensor = torch.Tensor
```
AnirudhDagar, line 41: @mli Not a good idea as I mentioned my concerns in the above comments

PR #1153, opened 2020-07-10
revert all d2l alias before evaluation

mli:
*Description of changes:*

For all alias such as `d2l.ones`, `d2l.reshape`, we will convert them to the framework native operators before evaluation. 

choice.

### DIFF
--- a/chapter_preface/index.md
+++ b/chapter_preface/index.md
@@ -361,6 +361,14 @@ import numpy as np
 import tensorflow as tf
 ```
 
+Note that the comment `#@save` is a special mark where the following function,
+class, or statements are saved in the `d2l` package.
+In the rest of the book,
+we often define more complicated functions or classes.
+For those that can be used later,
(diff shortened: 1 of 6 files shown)


mli:
Job d2l-en/PR-1153/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1153/

astonzhang:
one risk for replacing d2l.reshape to mxnet reshape is that it may increase number of characters in a line (we require each line of code has <=78 chars), which may cause undesirable line breaks due to page width in pdf

mli:
Job d2l-en/PR-1153/3 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1153/

mli:
it should not happen, as we change from `d2l.reshape(a, ...)` to `a.reshape(...)`. but what could happen is from `d2l.ones` to `torch.ones`, which increase 2 chars.

PR #1158, opened 2020-07-13
Change True to False in ResNet logic

Okroshiashvili:
I believe there should be ```False```. Otherwise, logic is not consistent.

choice.


mli:
Job d2l-en/PR-1158/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1158/

Issue #1150, opened 2020-07-10
MXNet/PT Non Ideal Results NiN


AnirudhDagar:
The loss/accuracy plots are problematic for both MXNet and PyTorch.

@mli @astonzhang 
http://preview.d2l.ai/d2l-en/master/chapter_convolutional-modern/nin.html

StevenJokes:
I said that they were caused by #1141:https://github.com/d2l-ai/d2l-en/issues/1141. Why did nobody fix it?

astonzhang:
@StevenJokes  plz see my comments https://github.com/d2l-ai/d2l-en/pull/1176

Issue #1141, opened 2020-07-08
[pytorch]Some questions in torch.py.


StevenJokes:
https://github.com/d2l-ai/d2l-en/commit/7a6d2973f85274f2e11cb6286bb68955254cc693#diff-7199ea3162d85ba4377898f52160727b

![image](https://user-images.githubusercontent.com/25657787/86955566-5d1c3a00-c18a-11ea-83ab-fb83a6f9dd3c.png)

![image](https://user-images.githubusercontent.com/25657787/86955734-a3719900-c18a-11ea-883b-08f38076a18d.png)

StevenJokes:
Please fix it quickly! @mli @astonzhang @AnirudhDagar

StevenJokes:
It also happens in mxnet.py and tensorflow.py, so it maybe influences all eval process.
I reinstall my system, and lost d2l old version. But I will not install d2l until it fixed.

StevenJokes:
prod!

StevenJokes:
No answer...

goldmermaid:
Hi @StevenJokes , please refer to the latest updates: https://github.com/d2l-ai/d2l-en/blob/master/d2l/torch.py.

Let us know if that solve your questions.

StevenJokes:
It didn't. 
![image](https://user-images.githubusercontent.com/25657787/87458505-acf07a80-c63c-11ea-8567-54333b170c3b.png)
in https://numpy.org/doc/stable/reference/generated/numpy.ndarray.size.html 
![image](https://user-images.githubusercontent.com/25657787/87458415-86324400-c63c-11ea-881d-1f789ac1c5ac.png)

StevenJokes:
The function was saved in torch.py, tensorflow.py, mxnet.py.
I have to PR first.

StevenJokes:
OK. PR all related.

astonzhang:
See my comments in https://github.com/d2l-ai/d2l-en/pull/1176

Issue #1123, opened 2020-07-06
Sudden drops in plots of loss/accuracy with extended training of a neural network


NishantTharani:
Following along with the [concise implementation of multiplayer perceptrons](https://d2l.ai/chapter_multilayer-perceptrons/mlp-concise.html) I then tried to train a neural network with one extra hidden layer, for 50 epochs instead of 10. The resulting plot of training loss/train acc/test acc exhibits sudden drops:

<img width="263" alt="image" src="https://user-images.githubusercontent.com/33641044/86604485-541a4580-bf9d-11ea-909b-76e8230948cf.png">

It does not always look like this - sometimes there are no drops and sometimes there is a drop and then a recovery followed by another drop, etc:

<img width="258" alt="image" src="https://user-images.githubusercontent.com/33641044/86604522-61373480-bf9d-11ea-941a-d112a6801ea8.png">

Comments by @AnirudhDagar on a [forum post](https://discuss.d2l.ai/t/concise-implementation-of-multilayer-perceptron/95/7) I made about this indicate that it could be an issue related to the plot function.

Here is a Jupyter notebook containing the code I ran: https://github.com/NishantTharani/GitSharing/blob/master/concise_multilayer_perceptrons.ipynb

AnirudhDagar:
Hi @NishantTharani,
I was able to verify the issue and reproduce your results. The reason for the sudden drops is because the training loss becomes NaN after a few epochs.
You can fix this easily by using a smaller learning rate, which will keep this in control. A good default for SGD is lr=0.01 or 0.05.
Please try reporting back your results with these. :)

StevenJokes:
#1141, `sum` may cause the result wrong

astonzhang:
> @AnirudhDagar
> #1141, `sum` may cause the result wrong.

This result is obtained after one additional layer without re-tuning hyperparameters. How could `sum` cause this result? Could you please be clearer? Thanks.

astonzhang:
@StevenJokes see my comments in https://github.com/d2l-ai/d2l-en/pull/1176

AnirudhDagar:
I and @goldmermaid discussed about this issue and she suggested it is probably due to the loss becoming NaN. Later, I verified that with a reduced learning rate as can be seen in my last comment. This probably hints at exploding gradients issue. What do you think @astonzhang?
I don't understand what Steven is suggesting.

astonzhang:
[...]

PR #1176, opened 2020-07-14
Try to help #1123,#1141,#1150 in softmax-regression-scratch.md

StevenJokes:
*Description of changes:*

choice.

### DIFF
--- a/chapter_linear-networks/softmax-regression-scratch.md
+++ b/chapter_linear-networks/softmax-regression-scratch.md
@@ -360,7 +360,7 @@ def evaluate_accuracy(net, data_iter):  #@save
     """Compute the accuracy for a model on a dataset."""
     metric = Accumulator(2)  # No. of correct predictions, no. of predictions
     for _, (X, y) in enumerate(data_iter):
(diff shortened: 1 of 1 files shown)


StevenJokes:
please recheck.

mli:
Python doesn't have `prod`, maybe we can use `math.prod` here

StevenJokes:
Or we change it back to `y.numpy().size`?

StevenJokes:
I can't figure out how to `d2lbook build lib`. Please change the required code for us first ,and tell me how to do it next time.

astonzhang:
For all the usage of `d2l.evaluate_accuracy` and `d2l.train_ch3` (it uses d2l.evaluate_accuracy) throughout the book, `y` is a 1-D array, so `sum(y.shape)` gives the same value of `y.size` in mx, pt, and tf.

Since you mentioned that sum(y.shape) causes problems in  #1123,#1141,#1150, could you be a little more specific on why? Thanks.

StevenJokes:
Can you explain the purpose of  "metric.add(accuracy(net(X), y), d2l.size(y))" in details for me?
It is hard to understand. A picture will be best.

PR #1289, opened 2020-08-04
backprop.md: Ex2: Don't require bias in the regularization term

yoderj:
Given the ambiguity of whether biases are included[1], I think it is good to specify whether they are needed not. I chose not in this suggested edit to keep the problem simpler.  Choosing to explicitly require it might give students practice in thinking about variables that depend on multiple paths, but my guess is they'll get that from reworking the weights anyway.

[1] "Whether we include a corresponding bias penalty  b2  can vary across implementations, and may vary across layers of a neural network. Often, we do not regularize the bias term of a network’s output layer." in section [4.5.1](https://d2l.ai/chapter_multilayer-perceptrons/weight-decay.html#norms-and-weight-decay)

*Description of changes:*

choice.


mli:
Job d2l-en/PR-1289/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1289/

PR #1344, opened 2020-08-13
 backprop.md: Make vector y boldface in markdown and forward.svg

yoderj:
Given that the output $\mathbf{o}$ is a vector, it seems likely that the label $y$ would be a vector, too.  (Unless, of course, the loss function is converting a scalar label into a one-shot label -- but in that case, I would expect a softmax output layer.)

I edited all $.*y instances in backprop.md and updated the font in forward.svg. Inscape really changed the format of the xml significantly, but it looks the same. I can manually edit the SVG to bring it closer to the original format if that is desirable.

choice.


mli:
Job d2l-en/PR-1344/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1344/

astonzhang:
For categories, $y$ is an integer, not a vector. It is the operator's concern to manipulate over $\mathbf{o}$ and $y$.

astonzhang:
Closing it for now. Just comment or reopen it if there are more questions.

PR #1345, opened 2020-08-13
Update probability.md

prateekvyas1996:
The numpy of mxnet not working, it's giving weird results. Replacing mxnet numpy with normal numpy working fine

*Description of changes:*
The numpy of mxnet is not giving exact results as normal numpy checked the d2l book code on my own system, just replacing the 'from mxnet import np,npx' to 'import numpy as np' giving exact results. There something wrong in the code please check

choice.

### DIFF
--- a/chapter_preliminaries/probability.md
+++ b/chapter_preliminaries/probability.md
@@ -44,9 +44,8 @@ To start, let us import the necessary packages.
 ```{.python .input}
 %matplotlib inline
 from d2l import mxnet as d2l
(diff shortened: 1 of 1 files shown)


astonzhang:
The entire D2L is based on mxnet.numpy rather than the official numpy.
If you believe the weird results are due to mxnet.numpy, please report the issue on https://github.com/apache/incubator-mxnet/issues. If they are caused by the book, you may reopen the PR to fix or report on our issue.

PR #1361, opened 2020-08-15
pytorch 10.3 transformer

ypandya:
*Description of changes:*

choice.

### DIFF
--- a/chapter_attention-mechanisms/transformer.md
+++ b/chapter_attention-mechanisms/transformer.md
@@ -30,7 +30,7 @@ On the flip side, Transformer differs from the seq2seq with attention model in t
 
 In the rest of this section, we will equip you with each new component introduced by Transformer, and get you up and running to construct a machine translation model.
 
-```{.python .input  n=1}
+```{.python .input}
 from d2l import mxnet as d2l
 import math
 from mxnet import autograd, np, npx
@@ -38,7 +38,7 @@ from mxnet.gluon import nn
 npx.set_np()
 ```
 
-```{.python .input  n=1}
+```{.python .input}
 #@tab pytorch
 from d2l import torch as d2l
 import math
@@ -78,7 +78,7 @@ $$\mathbf o = \mathbf W_o \begin{bmatrix}\mathbf o^{(1)}\\\vdots\\\mathbf o^{(h)
 
 Now we can implement the multi-head attention. Assume that the multi-head attention contain the number heads `num_heads` $=h$, the hidden size `num_hiddens` $=p_q=p_k=p_v$ are the same for the query,  key, and value dense layers. In addition, since the multi-head attention keeps the same dimensionality between its input and its output, we have the output feature size $d_o =$ `num_hiddens` as well.
 
-```{.python .input  n=2}
+```{.python .input}
 #@save
 class MultiHeadAttention(nn.Block):
     def __init__(self, num_hiddens, num_heads, dropout, use_bias=False, **kwargs):
@@ -119,7 +119,7 @@ class MultiHeadAttention(nn.Block):
         return self.W_o(output_concat)
 ```
 
-```{.python .input  n=2}
+```{.python .input}
 #@tab pytorch
 #@save
 class MultiHeadAttention(nn.Module):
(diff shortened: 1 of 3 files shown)


chapter_preliminaries/ndarray.md
```diff
@@ -55,7 +55,7 @@ To start, we import `tensorflow`. As the name is a little long, we often import
 it with a short alias `tf`.
 :end_tab:
 
-```{.python .input}
+```python
```
AnirudhDagar, line 58: Why change this file? I guess this has been mistakenly committed?

ypandya:
Thanks for rectifying that mistake. ndarray.md was not supposed to be part of the commit. One of the cells is timing out now. I'll check and see what I can do about this.

chapter_attention-mechanisms/transformer.md
```diff
@@ -436,11 +738,34 @@ model = d2l.EncoderDecoder(encoder, decoder)
...
+src_vocab, tgt_vocab, train_iter = d2l.load_data_nmt(batch_size, num_steps)
+
+encoder = TransformerEncoder(
+    len(src_vocab), key_size, query_size, value_size, num_hiddens,
+    norm_shape, ffn_num_input, ffn_num_hiddens, num_heads,
+    num_layers, dropout)
+decoder = TransformerDecoder(
+    len(src_vocab), key_size, query_size, value_size, num_hiddens,
+    norm_shape, ffn_num_input, ffn_num_hiddens, num_heads,
+    dropout)
```
ypandya, line 758:
```suggestion
    num_layers, dropout)
```

chapter_attention-mechanisms/transformer.md
```diff
@@ -414,6 +687,35 @@ class TransformerDecoder(d2l.Decoder):
         return self.dense(X), state
 ```
 
+```{.python .input}
+#@tab pytorch
+class TransformerDecoder(d2l.Decoder):
+    def __init__(self, vocab_size, key_size, query_size, value_size,
+                 num_hiddens, norm_shape, ffn_num_input, ffn_num_hiddens,
+                 num_heads, dropout, **kwargs):
```
ypandya, line 695:
```suggestion
                 num_heads, num_layers, dropout, **kwargs):
```

ypandya:
The final training block for pytorch is taking 10x the time taken by the mxnet code. I've not been able to find the root cause this yet. There is not any significant differences in the implementation. The major difference in pytorch and mxnet implementation is that the input values for each linear layer needs to be provided explicitly for pytorch.

AnirudhDagar:
@ypandya any updates on this?

ypandya:
> @ypandya any updates on this?

I have matched the mxnet and pytorch models in detail. There is no difference in the architecture. My early guess was that there maybe difference in the layer norm implementation in pytorch and mxnet. But the training takes the same amount of time even when I experimented with replacing layernorm with batchnorm. I could not find any reason for the difference which can cause the torch training to take 10x the time taken by mxnet. It would be really helpful if someone with more knowledge of mxnet can take a look at this.

AnirudhDagar:
Thanks! I see, I'll try to look at the PyTorch code if I can find the difference.
Maybe @astonzhang or @mli can have a look to spot the problem and to find a fix.

StevenJokes:
cuda is slower than cpu??
image

AnirudhDagar:
@ypandya I was able to resolve the issue and traced it back to a function defined in `seq2seq`  which needed some optimization. The implementation now is 60 fold faster than what it was before and it should pass CI as expected.

The whole problem was due to an inbuilt function which is present in MXNet and is missing in PyTorch. `sequence_mask` For now we have a good custom implementation for same, but I feel we can improve the code in this section in later revisions for a more fluent PyTorch Adaptation.

@astonzhang please take a look at the PR and review/merge. I think we should be good to go for now. Though I'll try to come up with a more elegant solution later.

AnirudhDagar (approved):

mli:
Job d2l-en/PR-1361/5 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1361/

astonzhang:
Thanks @ypandya and @AnirudhDagar!

PR #1406, opened 2020-08-27
Handles an edge case in 3.6.5 Accuracy

tinkuge:
In the definition of `accuracy`, `if len(y_hat.shape) >= 1:` would handle the case of `y_hat` containing prediction for only one example. Checking the second dimension of `y_hat.shape[1] > 1` is also redundant since `tf.argmax` would handle the case of tensor having only one element by returning the index `0`.

*Description of changes:*

choice.


mli:
Job d2l-en/PR-1406/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1406/

astonzhang:
Thanks. For the use case in D2L, if y_hat is just 1-D, no matter whether y_hat.shape[0] >= 1, it contains the predicted category index and we will simply compare y_hat with y.

PR #1423, opened 2020-09-04
[WIP][TF]Add GAN

StevenJokes:
https://github.com/StevenJokes/d2l-en-read/blob/master/chapter-generative-adversarial-networks/gan-tf1.ipynb


StevenJokes:
Help me if you would like and can...
@terrytangyuan

StevenJokes:
Why is my loss of generator 0? Why is loss of discriminator bigger than generator
(https://stackoverflow.com/questions/63763835/why-is-my-loss-of-generator-0-why-is-loss-of-discriminator-bigger-than-generato)

PR #1439, opened 2020-09-13
Spell Corrections and equations corrections

LakshKD:
*Description of changes:*

1. Correction of word "eastimating" -> "estimating"

2. According to the equations of NeuMF written on this page (16.6.2), the y_hat term should have z^(L-1) instead of z^(L) as the input to the Last layer of MLP network will be z^(L-1) and phi^(L) is the function that is getting applied to get the output from Last Layer, i.e., L. 

3. Accordingly equation 16.6.3 should also have the similar modification, I have changed that as well.

Thanks  

choice.


mli:
Job d2l-en/PR-1439/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1439/

astonzhang:
@cheungdaven PTAL

cheungdaven:
I think it ok to merge it. @LakshKD Thank you for correction.

PR #1441, opened 2020-09-14
[PyTorch] Add comments to train_epoch_ch3

AnirudhDagar:
*Description of changes:*
This clearly explains the two different cases which can take place in the training loop.
1. When using native optimizer and loss from PyTorch.
2. When we define our own updater and custom loss as above.

choice.


mli:
Job d2l-en/PR-1441/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1441/

PR #1464, opened 2020-09-24
Erroneous question statement

goldmermaid:
Restate the question based at the feedback from discussion forum.


mli:
Job d2l-en/PR-1464/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1464/

astonzhang:
We intended to check the understanding of "independence" between trials. This edit has changed the meaning.

goldmermaid:
Is we run the first test the second time, the two time test results should be independent, right?

astonzhang:
Since we assume conditional independence for the two test methods, we can have:
𝑃(𝐷1=1,𝐷2=1∣𝐻=1) = 𝑃(𝐷1=1∣𝐻=1)𝑃(𝐷2=1∣𝐻=1)
However, for the same test method, we cannot assume conditional independence. Two results of the same test method should be correlated.

goldmermaid:
The original question states that "run the first test a second time", so it is common for students to think of running two independent "first test". I have seen 3 students have the same [confusions](https://discuss.d2l.ai/t/probability/36/10).

astonzhang:
Yes, running the first test a second time means getting two results of the same test method.

astonzhang:
Just added some clarifications: https://github.com/d2l-ai/d2l-en/commit/ea4e69ec7b36f35804b9d2f608c928cdb0e65e28

PR #1475, opened 2020-09-29
amend lower bound

cuongvng:
*Description of changes:*

18.5.5 define F(x) with lower bound **0**:
image

However, 18.5.10 plugs in F(x) with lower bound **-inf**, which I found confusing.

image

choice.


mli:
Job d2l-en/PR-1475/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1475/

goldmermaid:
Hi @cuongvng , thanks for contributing! The 18.5.5 function measures the area between 0 and 𝑥. However, this is the special situation and we assume f is defined on [0, +inf]. In general, a function f(x) can be defined on any range in between [-inf, +inf]. 

Please closed this PR if it makes sense to you.

PR #1507, opened 2020-11-04
Tensorflow not available yet for Python 3.9

steinsag:
When creating conda env, it is important to enforce Python 3.8 as Tensorflow packages are not available yet for Python 3.9.

In addition, it is important to always specify `python`. Otherwise, miniconda will create an empty environment without pip and python. There is no need to upgrade pip afterwards as it will be latest version already.


mli:
Job d2l-en/PR-1507/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1507/

PR #1539, opened 2020-11-25
[TF] Bug in MultiFactorScheduler implementation Fix #1306

### LINKED ISSUE #1306: Bug in MultiFactorScheduler implementation TF
In [11.11.3.2. Multi Factor Scheduler](http://preview.d2l.ai/d2l-en/master/chapter_optimization/lr-scheduler.html#multi-factor-scheduler) the `__call__` in the class defined will behave incorrectly when the user inputs an argument where size of list in `step` is more than 2.

```python
class MultiFactorScheduler:
  def __init__(self, step, factor, base_lr):
    self.step = step
    self.factor = factor
    self.base_lr = base_lr

  def __call__(self, epoch):
    if epoch in range(self.step[0], (self.step[1] + 1)):
      return self.base_lr * self.factor
    else:
      return self.base_lr

scheduler = MultiFactorScheduler(step=[15, 30], factor=0.5, base_lr=0.5)
d2l.plot(tf.range(num_epochs),
         [scheduler(t) for t in range(num_epochs)])
```

Eg if step = [10,15,30] the TF implementation is not correct for such case. 

PTAL @abhinavsp0730 @terrytangyuan 
cc @astonzhang

abhinavsp0730:
*Description of changes:*

choice.

Bug Fix: #1306

@AnirudhDagar @astonzhang @terrytangyuan, please review 😃 .

### DIFF
--- a/chapter_optimization/lr-scheduler.md
+++ b/chapter_optimization/lr-scheduler.md
@@ -337,8 +337,9 @@ class MultiFactorScheduler:
         self.base_lr = base_lr
   
     def __call__(self, epoch):
-        if epoch in range(self.step[0], (self.step[1] + 1)):
-            return self.base_lr * self.factor
+        if epoch in self.step:
+            self.base_lr = self.base_lr * self.factor
+            return self.base_lr
         else:
             return self.base_lr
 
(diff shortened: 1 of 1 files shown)


mli:
Job d2l-en/PR-1539/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1539/

terrytangyuan (approved):

PR #1542, opened 2020-11-26
Unnecessary: Transposing a 1-rank tensor

StevenJokess:
https://discuss.d2l.ai/t/sequence-models/114/8

*Description of changes:*

choice.


mli:
Job d2l-en/PR-1542/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1542/

DOH-Manada:
The book doesn't state this, but you're not setting each row of the matrix as a set of Tau observations, you're actually setting each column equal to the sequence values as a column vector of size (len(vector) - Tau)//step_size, i.e., for each sequence of values y_ti = [i :: step_size] for i = 0, ... Tau-1.

```
matrix = [ y_t0.T | ... | y_tTau-1.T | y_tTau.T ]
```
in which the last column vector is the set of labels.

PR #1589, opened 2020-12-29
Elaborate the loss explanation

adriaant:
The rationale for the loss function calculation uses a good example, but the explanation for using indexing could be a bit more elaborate, imho.

*Description of changes:*

choice.


mli:
Job d2l-en/PR-1589/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1589/

astonzhang:
Thanks and I'll further revise. Could you replace your ID with name in the acknowledgements:

PR #1591, opened 2020-12-29
Fix moving var initialization: 0 -> 1

vfdev-5:
*Description of changes:*

By default frameworks like PyTorch/Tensorflow initialize moving variance to ones instead of zeros (here):
- pytorch: https://github.com/pytorch/pytorch/blob/a111a9291c704c218cae99aeb3d8be23a54368ff/torch/nn/modules/batchnorm.py#L60
- tensorflow: https://www.tensorflow.org/api_docs/python/tf/keras/layers/BatchNormalization
- maybe, same for MXNet

In this PR, simply replaces `zeros` to `ones` for moving variance initialization.

choice.


mli:
Job d2l-en/PR-1591/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1591/

PR #1623, opened 2021-01-10
Include difference of net(X) and net.forward(X)

floriandonhauser:
In pytorch, `net(X)` is shorthand for `net.__call__(X)` which is not exactly the same as `net.forward(X)`.
`net.__call__(X)` also includes hooks which might be necessary for the forward and backward pass.
This book, as far as I understand, does not include any mention of Pytorch's hooks and the reader might think that both calls are identical.
However, `net(X)` is the right way to use the model, since calling `net.forward(X)` directly will not call possibly required hooks.
In my opinion, this difference should be mentioned at some point in the book.
The exact wording can be changed, of course, but some explanation of the difference would be helpful for the reader.

This difference is also present in MXNet and hooks should be explained here as well.

[Source code in Pytorch](https://github.com/pytorch/pytorch/blob/9d8bd216f9e29251685389a22dce69614ab648a5/torch/nn/modules/module.py#L908)
[Source code in MXNet](https://github.com/apache/incubator-mxnet/blob/124d8417984ed9205f972eb1c2cadbf028b94eb3/python/mxnet/gluon/block.py#L846)

choice.

### DIFF
--- a/chapter_deep-learning-computation/model-construction.md
+++ b/chapter_deep-learning-computation/model-construction.md
@@ -156,9 +156,14 @@ it chains each `Block` in the list together,
 passing the output of each as the input to the next.
 Note that until now, we have been invoking our models
 via the construction `net(X)` to obtain their outputs.
-This is actually just shorthand for `net.forward(X)`,
+This is actually just shorthand for `net.__call__(X)`,
 a slick Python trick achieved via
 the `Block` class's `__call__` function.
+This `__call__`function then uses net.forward(X) and also
+executes any hooks of the model, which are functions the model requires
+to be executed during the forward or backward pass.
+In general, a model should be used with `net(X)` and not
+`net.forward(X)`, so all hooks are executed correctly.
 :end_tab:
 
 :begin_tab:`pytorch`
--- a/chapter_deep-learning-computation/model-construction.md
+++ b/chapter_deep-learning-computation/model-construction.md
@@ -186,7 +186,7 @@ the `Module` class's `__call__` function.
 This `__call__`function then uses net.forward(X) and also
 executes any hooks of the model, which are functions the model requires
 to be executed during the forward or backward pass.
-In general, a Module should be used with `net(X)` and not
+In general, a `Module` should be used with `net(X)` and not
 `net.forward(X)`, so all hooks are executed correctly.
 :end_tab:
 
--- a/chapter_deep-learning-computation/model-construction.md
+++ b/chapter_deep-learning-computation/model-construction.md
@@ -161,8 +161,8 @@ a slick Python trick achieved via
 the `Block` class's `__call__` function.
 This `__call__`function then uses net.forward(X) and also
 executes any hooks of the model, which are functions the model requires
-to be executed during the forward or backward pass.
-In general, a model should be used with `net(X)` and not
+to be executed during the forward pass.
+In general, a Block should be used with `net(X)` and not
 `net.forward(X)`, so all hooks are executed correctly.
 :end_tab:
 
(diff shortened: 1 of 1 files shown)


mli:
Job d2l-en/PR-1623/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1623/

astonzhang:
Thanks. If you want to update text about PyTorch, do you wish to modify the text following the :begin_tab:`pytorch` ? (This PR so far modifies text following the :begin_tab:`mxnet` )

floriandonhauser:
Actually, for mxnet, you also have hooks. See: [the source code of Block](https://github.com/apache/incubator-mxnet/blob/124d8417984ed9205f972eb1c2cadbf028b94eb3/python/mxnet/gluon/block.py#L846)
I would suggest both to change both tabs, mxnet and Pytorch.

mli:
Job d2l-en/PR-1623/2 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1623/

floriandonhauser:
For TensorFlow Keras I am a bit unsure what the exact difference is between call() and __call__().
[The sourcecode](https://github.com/tensorflow/tensorflow/blob/287ab9046604c277908969e20b7b4c0febe8b4ab/tensorflow/python/keras/engine/training.py#L465) says
"Note: This method should not be called directly. It is only meant to be
    overridden when subclassing `tf.keras.Model`.
    To call a model on an input, always use the `__call__` method,
    i.e. `model(inputs)`, which relies on the underlying `call` method."
Therefore, this should probably also be included in the book so readers do not use call(X) because they think it is exactly the same.

astonzhang:
Thanks. @AnirudhDagar Could you review the PyTorch part?

chapter_deep-learning-computation/model-construction.md
```diff
@@ -156,9 +156,14 @@ it chains each `Block` in the list together,
...
 Note that until now, we have been invoking our models
 via the construction `net(X)` to obtain their outputs.
-This is actually just shorthand for `net.forward(X)`,
+This is actually just shorthand for `net.__call__(X)`,
 a slick Python trick achieved via
 the `Block` class's `__call__` function.
+This `__call__`function then uses net.forward(X) and also
+executes any hooks of the model, which are functions the model requires
+to be executed during the forward pass.
+In general, a Block should be used with `net(X)` and not
```
AnirudhDagar, line 165:
```suggestion
In general, a `Block` should be used with `net(X)` and not
```

chapter_deep-learning-computation/model-construction.md
```diff
@@ -175,9 +180,14 @@ it chains each block in the list together,
...
 via the construction `net(X)` to obtain their outputs.
-This is actually just shorthand for `net.forward(X)`,
+This is actually just shorthand for `net.__call__(X)`,
 a slick Python trick achieved via
-the Block class's `__call__` function.
+the `Module` class's `__call__` function.
+This `__call__`function then uses net.forward(X) and also
+executes any hooks of the model, which are functions the model requires
+to be executed during the forward or backward pass.
+In general, a Module should be used with `net(X)` and not
```
AnirudhDagar, line 189:
```suggestion
In general, a `Module` should be used with `net(X)` and not
```

AnirudhDagar (changes requested):
Overall looks good to me. Thanks for this handy addition. Please make the suggested change.

AnirudhDagar (approved):

mli:
Job d2l-en/PR-1623/3 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1623/

mli:
Job d2l-en/PR-1623/4 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1623/

AnirudhDagar:
@astonzhang LGTM!

astonzhang:
Thanks. I'll merge and edit the MXNet part

floriandonhauser:
@astonzhang The MXNet has actually been changed as well. The missing one is TensorFlow for which @terrytangyuan is responsible as far as I know. As I wrote above, a similar difference between `__call__()` and `call()` exists here as well.

PR #1660, opened 2021-02-12
bug fix on batch norm

nikisix:
There seems to be a discrepancy regarding the LeNet specified in 7.4 vs the one specified in 6.6 –
the LeNet in 7.4 uses MaxPooling, while the original uses AvgPooling.

You want to control for pooling while considering Batch Normalization’s effect, no?

Furthermore simply changing the original LeNet from Avg -> Max Pooling yields an accuracy increase from:
```
(Avg) loss 0.467, train acc 0.824, test acc 0.797
->
(Max) loss 0.411, train acc 0.849, test acc 0.841
```

Versus the BatchNormalization scores reported in chapter 7.4 hovering around:
```
loss 0.250, train acc 0.909, test acc 0.849
```
It would seem most of the effect is explained by the choice of Pooling rather than the application of Batch Normalization.

choice.


mli:
Job d2l-en/PR-1660/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1660/

PR #1672, opened 2021-03-12
[TF]Add get_params as a parameter for train_ch8 for tensorflow

floriandonhauser:
If we want to save train_ch8 in tensorflow.py, we need access to get_params.
Since this method is not saved and will be different for e.g. lstm, we need to pass it in as a parameter.
Changing train_ch8 to d2l.train_ch8 for rnn-scratch (tensorflow version) will currently cause an error!
I also changed rnn-scratch where this method is defined.
I checked this new version of rnn-scratch and it runs as expected.

choice.


mli:
Job d2l-en/PR-1672/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1672/

terrytangyuan (approved):

Issue #1678, opened 2021-03-17
The second training with use_random_iter=True in rnn-scratch does not reset the weights


floriandonhauser:
I am currently working on chapters 8 and 9 for the tensorflow version (already merged 8.6).
I have found that the last code block in 8.5 continues the training of the previous code block.
In my opinion, the training should begin from new.

![rnn-scratch-training](https://user-images.githubusercontent.com/35968719/111464106-aa40aa80-8720-11eb-9e5a-56c272465813.png)

The new code to do so would be about 2 lines longer each:
```{.python .input}
#@tab mxnet
net = RNNModelScratch(len(vocab), num_hiddens, d2l.try_gpu(), get_params,
                      init_rnn_state, rnn)
train_ch8(net, train_iter, vocab, lr, num_epochs, d2l.try_gpu(),
          use_random_iter=True)
```

```{.python .input}
#@tab pytorch
net = RNNModelScratch(len(vocab), num_hiddens, d2l.try_gpu(), get_params,
                      init_rnn_state, rnn)
train_ch8(net, train_iter, vocab, lr, num_epochs, d2l.try_gpu(),
          use_random_iter=True)
```

```{.python .input}
#@tab tensorflow
with strategy.scope():
    net = RNNModelScratch(len(vocab), num_hiddens, init_rnn_state, rnn, get_params)

train_ch8(net, train_iter, vocab_random_iter, num_hiddens, lr,
          num_epochs, strategy, use_random_iter=True)
```

@terrytangyuan  Which version is correct and should be used for the TF version I am working on?
Do we want to restart with use_random_iter=True or continue the training?
The whole TensorFlow training loop has a problem where new models will not work with it if they have different parameters.
The reason is that get_params is called to early which only works if we have used the model (call) before training. Otherwise, the params of the model have not yet been initialized!
I am working on a fix for this as well and will create a PR for 9.1 (GRU) and 9.2 (LSTM) with it most likely tomorrow.

terrytangyuan:
Let's stick with the behavior in MXNet/PyTorch implementations for now to be consistent.

floriandonhauser:
I think consistency is important but I suggest we change all three variants in the near future.
Other people have also found this problem and have mentioned it in the discussion of this chapter:
[https://discuss.d2l.ai/t/implementation-of-recurrent-neural-networks-from-scratch/486/9](https://discuss.d2l.ai/t/implementation-of-recurrent-neural-networks-from-scratch/486/9)

[...]

PR #1680, opened 2021-03-17
[TF] Add TensorFlow implementation of GRU and make training loop more flexible

floriandonhauser:
This PR adds a TensorFlow implementation for GRU which is somewhat based on the code of @PHOENIXFURY007  and his PR #1597
The code contains both the implementation from scratch and the concise one with keras.

Some changes had to be made to the training loop and the models from the previous chapters:
- When we initialize RNNModelScratch, we now pass in the function get_params which is executed at initialization to create the learnable weights.
- The training loop does not access these weights before the first model invocation. If we access them to early before invocation, keras implementations will return an empty list. The weights for keras are created during the first model invocation (call).
- The weights of the RNN (and later RNN models like GRU or LSTM) are accessed using model.trainable_variables for the implementation from scratch since this makes it compatible with Keras models. **This decision is up for discussion and we could use an if-else statement depending on the type of the model.**
- Parameters are now never passed in as arguments in addition to the model but always taken from the model using model.trainable_variables. This solves the issue of accessing them too early or having wrong values in params.
- The RNNModel now accepts more than two output values from the keras layer. The first output is always Y and the rest are a saved together as the list state. For RNN and GRU, the list state is just the state value. However for keras implementation for LSTM, we get back 3 values! H and C are then combined into state as is the case for the other frameworks. The code, hence, is compatible with more different RNN layers. I also experimented with adding a lambda layer in between, but this causes too much added complexity in my opinion. **However, I am always thankful for other suggestions.**

I already have a finished LSTM implementation in TensorFlow which I will create a PR for, as soon as this PR is merged.

choice.

### DIFF
--- a/chapter_recurrent-modern/gru.md
+++ b/chapter_recurrent-modern/gru.md
@@ -183,6 +183,15 @@ batch_size, num_steps = 32, 35
 train_iter, vocab = d2l.load_data_time_machine(batch_size, num_steps)
 ```
 
+```{.python .input}
+#@tab tensorflow
+from d2l import tensorflow as d2l
+import tensorflow as tf
+
+batch_size, num_steps = 32, 35
+train_iter, vocab = d2l.load_data_time_machine(batch_size, num_steps)
+```
+
 ### Initializing Model Parameters
 
 The next step is to initialize the model parameters.
@@ -242,6 +251,29 @@ def get_params(vocab_size, num_hiddens, device):
     return params
 ```
 
+```{.python .input}
+#@tab tensorflow
+def get_params(vocab_size, num_hiddens):
+    num_inputs = num_outputs = vocab_size
+    
+    def normal(shape):
+        return d2l.normal(shape=shape,stddev=0.01,mean=0,dtype=tf.float32)
+
+    def three():
+        return tf.Variable(normal((num_inputs, num_hiddens)), dtype=tf.float32),\
+               tf.Variable(normal((num_hiddens, num_hiddens)), dtype=tf.float32),\
+               tf.Variable(d2l.zeros(num_hiddens), dtype=tf.float32)
+    
+    W_xz, W_hz, b_z = three()  # Update gate parameters
+    W_xr, W_hr, b_r = three()  # Reset gate parameters
+    W_xh, W_hh, b_h = three()  # Candidate hidden state parameters
+    # Output layer parameters
+    W_hq = tf.Variable(normal((num_hiddens, num_outputs)), dtype=tf.float32)
+    b_q = tf.Variable(d2l.zeros(num_outputs), dtype=tf.float32)
+    params = [W_xz, W_hz, b_z, W_xr, W_hr, b_r, W_xh, W_hh, b_h, W_hq, b_q]
+    return params
+```
+
 ### Defining the Model
 
 Now we will define the hidden state initialization function `init_gru_state`. Just like the `init_rnn_state` function defined in :numref:`sec_rnn_scratch`, this function returns a tensor with a shape (batch size, number of hidden units) whose values are all zeros.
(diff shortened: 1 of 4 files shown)


mli:
Job d2l-en/PR-1680/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1680/

chapter_recurrent-modern/gru.md
```diff
@@ -242,6 +251,29 @@ def get_params(vocab_size, num_hiddens, device):
...
+```{.python .input}
+#@tab tensorflow
+def get_params(vocab_size, num_hiddens):
+    num_inputs = num_outputs = vocab_size
+    
+    def normal(shape):
+        return d2l.normal(shape=shape,stddev=0.01,mean=0,dtype=tf.float32)
+
+    def three():
+        return tf.Variable(normal((num_inputs, num_hiddens)), dtype=tf.float32),\
```
terrytangyuan, line 263: Use parenthesis to surround these and avoid the use of `\`

chapter_recurrent-neural-networks/rnn-concise.md
```diff
@@ -235,11 +235,13 @@ class RNNModel(tf.keras.layers.Layer):
         self.rnn = rnn_layer
         self.vocab_size = vocab_size
         self.dense = tf.keras.layers.Dense(vocab_size)
-    def call(self, inputs, state, _):
+
+    def call(self, inputs, state):
         X = tf.one_hot(tf.transpose(inputs), self.vocab_size)
-        Y,state = self.rnn(X, state)
+        Y, *state = self.rnn(X, state)
```
terrytangyuan, line 241: Perhaps add a comment here on why the wildcard is needed?

terrytangyuan (changes requested):

chapter_recurrent-modern/gru.md
floriandonhauser: Done

chapter_recurrent-neural-networks/rnn-concise.md
floriandonhauser: Done

mli:
Job d2l-en/PR-1680/2 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1680/

terrytangyuan (approved):

chapter_recurrent-neural-networks/rnn-scratch.md
```diff
@@ -412,16 +411,16 @@ def predict_ch8(prefix, num_preds, net, vocab, device):  #@save
 
 ```{.python .input}
 #@tab tensorflow
-def predict_ch8(prefix, num_preds, net, vocab, params):  #@save
+def predict_ch8(prefix, num_preds, net, vocab): #@save
```
astonzhang, line 414: two space before an inline comments

chapter_recurrent-neural-networks/rnn-scratch.md
```diff
@@ -732,20 +731,17 @@ def train_ch8(net, train_iter, vocab, lr, num_epochs, device,
 ```{.python .input}
 #@tab tensorflow
 #@save
-def train_ch8(net, train_iter, vocab, num_hiddens, lr, num_epochs, strategy,
-              get_params, use_random_iter=False):
+def train_ch8(net, train_iter, vocab, num_hiddens, lr, num_epochs, strategy, use_random_iter=False):
```
astonzhang, line 734: max 80 chars per line

astonzhang:
@floriandonhauser Thanks! I just left some minor comments. Besides, do you want to fix https://github.com/d2l-ai/d2l-en/issues/1678 after this?

floriandonhauser:
> @floriandonhauser Thanks! I just left some minor comments. Besides, do you want to fix #1678 after this?

Thank you for the feedback.
I will create a PR for issue #1678 after I am done with more RNN chapters. I'm currently working on LSTM and will create a PR for it very soon™ .

PR #1693, opened 2021-03-23
Fix for issue #1678 and remove unused parameter for train_ch8 in TF

floriandonhauser:
This PR fixes issue #1678 and it also removes an unused parameter for train_ch8 in TensorFlow.
train_ch8 now has the same parameters and is used the same way in all 3 frameworks, except for the device in mxnet and PyTorch which is called strategy in TensorFlow (since they are not 100% the same thing).
The PR also includes some minor formatting fixes which were given as feedback to prior PR after they had been merged.

After this PR is merged, I will create a PR for my finished TensorFlow implementation of chapter 9.3.

choice.

### DIFF
--- a/chapter_recurrent-modern/gru.md
+++ b/chapter_recurrent-modern/gru.md
@@ -372,7 +372,7 @@ num_epochs, lr = 500, 1
 with strategy.scope():
     model = d2l.RNNModelScratch(len(vocab), num_hiddens, init_gru_state, gru, get_params)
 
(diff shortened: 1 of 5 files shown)


mli:
Job d2l-en/PR-1693/2 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1693/

terrytangyuan (approved):

PR #1743, opened 2021-05-12
enable pytorch pdf build

mli:
*Description of changes:*

choice.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -59,6 +59,11 @@ stage("Build and Publish") {
       conda activate ${ENV_NAME}
       d2lbook build pdf
       """
+      
+      sh label:"Build Pytorch PDF", script:"""set -ex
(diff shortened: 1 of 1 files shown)


AnirudhDagar:
I was going through the logs and I think the error with the PyTorch PDF build is due to an empty index file. See [this](https://github.com/d2l-ai/d2l-pytorch-sagemaker/blob/master/chapter_deep-learning-computation/index.ipynb) zero byte file for example. There are some sections of the book not recognised for PyTorch including all the chapter index files, as well as some sections like Deferred Init which are MXNet specific. Sections which do not have a PyTorch implementation should be skipped for the pdf build to work (or for the time being we can utilize the MXNet section in that case).

This will solve empty notebook issues like [these](https://github.com/d2l-ai/d2l-book/issues/18).

PR #1759, opened 2021-05-23
Modifying grad_clipping function of TF

biswajitsahoo1111:
The existing `grad_clipping` function won't work when gradients are `IndexedSlices`. When embedding layers are used in the model (as in section 9.7 and in chapter 10), gradients of embedding layer are stored as `tf.IndexedSlices`. So before applying gradient clipping, `IndexedSlices` need to be converted into tensors. The new function basically does that. It also works perfectly fine when all the gradients are tensors. So by modifying the old function we can make it useful for future chapters.

choice.

### DIFF
--- a/chapter_recurrent-neural-networks/rnn-scratch.md
+++ b/chapter_recurrent-neural-networks/rnn-scratch.md
@@ -520,19 +520,23 @@ def grad_clipping(net, theta):  #@save
 
 ```{.python .input}
 #@tab tensorflow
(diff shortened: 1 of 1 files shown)


mli:
Job d2l-en/PR-1759/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1759/

biswajitsahoo1111:
This has been superseded by [1768](https://github.com/d2l-ai/d2l-en/pull/1768). So closing.

PR #1797, opened 2021-06-15
Bump torch to 1.9.0 and torchvision to 0.10.0

AnirudhDagar:
*Description of changes:*
Recently PyTorch released new stable versions. We should bump our dependencies and check if the CI runs green with the new version. I'll remove `d2lbook clear` once the complete build works in the CI.

choice.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -24,6 +24,7 @@ stage("Build and Publish") {
 
       sh label: "Sanity Check", script: """set -ex
       conda activate ${ENV_NAME}
+      d2lbook clear
       d2lbook build outputcheck tabcheck
       """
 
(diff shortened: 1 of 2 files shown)


AnirudhDagar:
@astonzhang just found out that cuda 10.1 wheels are not published [here](https://download.pytorch.org/whl/torch_stable.html). Can we update the CI to CUDA 10.2?

astonzhang:
> @astonzhang just found out that cuda 10.1 wheels are not published [here](https://download.pytorch.org/whl/torch_stable.html). Can we update the CI to CUDA 10.2?

Currently we still depend on TF 2.3.1, which is incompatible with cu10.2:
https://www.tensorflow.org/install/source#gpu

I'll see if we upgrade CUDA to even higher versions after one or two more releases.

PR #1806, opened 2021-06-21
Update text-preprocessing.md

armingh2000:
change list to element, semantic problem

*Description of changes:*
I think there should be element instead of list. Because list doesn't make any sense and in that list (lines), elements are also not lists (they are strings). 
I changed this line:
The following tokenize function takes a list (lines) as the input, where each **_list_** is a text sequence (e.g., a text line).
to this:
The following tokenize function takes a list (lines) as the input, where each **_element_** is a text sequence (e.g., a text line).

choice.


mli:
Job d2l-en/PR-1806/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1806/

PR #1824, opened 2021-07-09
[PyTorch] Shared memory, PyTorch NumPy Bridge

AnirudhDagar:
*Description of changes:*
Fix text taken from `MXNet`. This is not true for `PyTorch`. We have shared memory for `NumPy-PyTorch` conversion.
Also, changed the way to switch from `NumPy array to `PyTorch tensor. `torch.from_numpy()` is the correct way as it will keep the `dtype` information.

@terrytangyuan could you also comment about the `TensorFlow` situation for this? We might need to fix `TensorFlow` text as well here.

choice.

### DIFF
--- a/chapter_preliminaries/ndarray.md
+++ b/chapter_preliminaries/ndarray.md
@@ -636,14 +636,22 @@ computation(X, Y)
 
 
 ## Conversion to Other Python Objects
-
-[**Converting to a NumPy tensor**], or vice versa, is easy.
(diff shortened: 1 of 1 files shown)


mli:
Job d2l-en/PR-1824/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1824/

terrytangyuan (commented):
This looks fine as is.

PR #1836, opened 2021-07-21
Fix the problem where tqdm bar flushes d2l.Animator

kss39:
*Description of changes:*
Hi Professors and developers,

I'm an incoming graduate student at University of Michigan, currently learning Dive into Deep Learning online on Bilibili. When I was trying to combine tqdm progress bar and d2l.Animator, I found that the progress bar constantly flushes out the matplotlib figure, making them impossible to be used together.

Here's a video explaining the hassle and how to fix it:
https://www.bilibili.com/video/BV12P4y147ye/
Sorry but it's in Chinese, I can make another English version if needed.
(Now there's a minimal piece of code below the reproduce the problem.)

choice.

### DIFF
--- a/d2l/mxnet.py
+++ b/d2l/mxnet.py
@@ -273,6 +273,7 @@ class Animator:
 
     def add(self, x, y):
         # Add multiple data points into the figure
+        display.clear_output(wait=True)
         if not hasattr(y, "__len__"):
             y = [y]
         n = len(y)
@@ -291,7 +292,6 @@ class Animator:
             self.axes[0].plot(x, y, fmt)
         self.config_axes()
         display.display(self.fig)
-        display.clear_output(wait=True)
 
 
 # Defined in file: ./chapter_linear-networks/softmax-regression-scratch.md
--- a/d2l/tensorflow.py
+++ b/d2l/tensorflow.py
@@ -284,6 +284,7 @@ class Animator:
 
     def add(self, x, y):
         # Add multiple data points into the figure
+        display.clear_output(wait=True)
         if not hasattr(y, "__len__"):
             y = [y]
         n = len(y)
@@ -302,7 +303,6 @@ class Animator:
             self.axes[0].plot(x, y, fmt)
         self.config_axes()
         display.display(self.fig)
-        display.clear_output(wait=True)
 
 
 # Defined in file: ./chapter_linear-networks/softmax-regression-scratch.md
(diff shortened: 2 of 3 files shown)


mli:
Job d2l-en/PR-1836/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1836/

kss39:
Some problem occured... it actually shows the result from last epoch, but not the current one (meaning that the last added data will be missing). Looks like this fix would not work. Sorry...

But is there anyway that we can make this animator compatible with tqdm? Thanks!

kss39:
I've also tried to use this:
```
        if not self.display:
            self.display = display.display(self.fig, display_id=True)
        else:
            self.display.update(self.fig)
```
to replace the original `display.display(self.fig); display.clear_output(wait=True)` part. But it will also print an additional figure when the cell finishes running.

kss39:
Maybe it's better if I create an issue instead of a pull request since I'm afraid that's out of my ability to fix this problem... sorry.

minimal code to reproduce the problem:
```
from tqdm.auto import tqdm
from d2l import torch as d2l
import time
import math
from IPython import display

%matplotlib inline

# A dummy function for train_acc
def sigmoid(x):
    return 1 / (1 + math.exp(-x))

epochs = 5
batch_count = 5
animator = Animator(xlabel='epoch', xlim=[1, epochs],
             ylim=[0.0, 1.0], legend=['train_acc'])
for i in range(epochs):
    for j in (range(batch_count)):
        time.sleep(0.3)
    animator.add(i + 1, (sigmoid(i / 5) - 0.5) * 2)
```

Versions: `d2l=0.16.6, tqdm=4.61.0`
The tqdm progress bar will flush the Animator figure away. My fix is effective, but it will print out an additional figure by the end of the running cell.

Thank you.

mli:
Job d2l-en/PR-1836/2 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1836/

goldmermaid:
Hi @kss39 , thanks for raising the issue. I suspect both `tqdm` and `IPython` return an object in the end, that's why you see two objects. If you can turn the `tqdm` output off, that may solve the issue.

astonzhang:
Thanks. I am closing it for now as we've been revising this function recently.

kss39:
Thank you and sorry, as I really didn't find a proper way to fix or improve this...

PR #1873, opened 2021-08-08
Removal of all ';' characters

BrianPulfer:
Removed non-useful ';' characters from python code cells

*Description of changes:*

choice.


terrytangyuan (commented):
This appears in most of the code cells, not just for this specific markdown file. We should probably decide on a style and then be consistent across the chapters.

PR #1875, opened 2021-08-09
Update machine-translation-and-dataset.md

tcmyxc:
if not set the encoding, a 'gbk' codec can't decode byte 0x93 error may appear on the win10 computer


tcmyxc:
Is it because I submitted it in the wrong format?

PR #1954, opened 2021-10-30
Fix sneaky Unicode symbols in math

nils-werner:
*Description of changes:*

The document has a few "sneaky Unicode symbols" that look like their proper counterpart, but in fact are not.

The symbols were causing LaTeX errors while trying out EPUB rendering (d2l-ai/d2l-book#37)


mli:
Job d2l-en/PR-1954/4 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1954/

PR #1967, opened 2021-11-18
fix function predict_seq2seq

Yeluoqianqiu:
`net.decoder.attention_weights` should be `net.decoder.attention_weights()`, attention_weights is a function, we should call it here, otherwise attention_weight_seq will be a list of functions, rather than a list of attention_weights. and when we exec `attention_weights = torch.cat([step[0][0][0] for step in dec_attention_weight_seq],0).reshape((1, 1, -1, num_steps))` later, an error will  occur: 'method' object is not subscriptable.

*Description of changes:*

choice.


mli:
Job d2l-en/PR-1967/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-1967/

PR #2009, opened 2022-01-14
[PyTorch] hotfix r0.17.2

AnirudhDagar:
*Description of changes:*
The following sections are fixed:

* The hotdog class in imagenet is number [934](https://gist.github.com/yrevar/942d3a0ac09ec9e5eb3a#file-imagenet1000_clsidx_to_labels-txt-L935) instead of 713.
* 1/2 is added back in mseloss for chapter_optimization in `train_concise_ch11` because the plots won't show since without the 1/2 the loss gets doubled and the ylim of the training curve has been fixed between 0.22 and 0.35. Without the 1/2 factor the loss never shows up on the plots.
* lr=0.01 makes learning smoother in the updated notebooks.
* l.mean() instead of l.sum() (in the weight-decay section) is more stable, producing the expected plots.

choice.

### DIFF
--- a/chapter_computer-vision/fine-tuning.md
+++ b/chapter_computer-vision/fine-tuning.md
@@ -362,7 +362,7 @@ hotdog_w.shape
 ```{.python .input}
 #@tab pytorch
 weight = pretrained_net.fc.weight
-hotdog_w = torch.split(weight.data, 1, dim=0)[713]
+hotdog_w = torch.split(weight.data, 1, dim=0)[934]
 hotdog_w.shape
 ```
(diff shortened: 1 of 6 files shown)


mli:
Job d2l-en/PR-2009/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-2009/

astonzhang:
Thanks. I'll move the comments `# Note: `MSELoss` computes squared error without the 1/2 factor` closer to the multiplication by 1/2 to improve readability.

PR #2010, opened 2022-01-16
Equation boundaries changes.

qingfengtommy:
I changed the lower boundary of entropy for continuous variables from  x to negative inf and add positive inf as the upper boundary.

*Description of changes:*

choice.


mli:
Job d2l-en/PR-2010/1 is complete. 
Check the results at http://preview.d2l.ai/d2l-en/PR-2010/

astonzhang:
Is it necessary？

qingfengtommy:
> Hi @qingfengtommy , thanks for your contribution! Can we know the necessity to include the infinity sighs? (From my understanding, sometimes x may not have value in the whole range.)

Sorry for late reply. Actually based on the mathematics textbooks it seems like to use the limit boundaries. But in deed it's also ok for using the x in the equation.

PR #2035, opened 2022-02-02
Clarify that weight decay is equivalent to l2 regularization in SGD

astonzhang:
Weight decay and L2 regularization may not be equivalent in non-SGD algorithms.

choice.


cheungdaven:
lgtm

cheungdaven (approved):

PR #2053, opened 2022-02-26
Include reference to total derivate in bptt.md

AlexSauer:
Equation (8.7.4) or line 118 is confusing at the moment because the left hand side and the first term on the right are exactly the same. The correct way would probably to change the notation completely to total derivatives (i.e. $\frac{d h_t}{d w_h}$) but I understand that this might be confusing for some readers as well and you don't want to confuse with too much different notation. Therefore, I think a good compromise would be to at least explicitly mention that the reason for the sum in equation (8.7.4) is the presence of the total derivative to give a potentially confused reader a hint what to google for.

*Description of changes:*

choice.


astonzhang:
@AlexSauer Thanks! Could you send another PR to replace your GitHub ID with your name in our acknowledgement:

PR #2058, opened 2022-03-07
[TF] Fix d2l API section

AnirudhDagar:
*Description of changes:*
Earlier `d2l.torch` was repeated twice.

choice.


astonzhang (approved):

AnirudhDagar:
Please don't merge yet.

AnirudhDagar:
This also fixes the other issue with the master branch d2l API section which only shows one framework even after changing tabs. I'm not sure if this solution is the most elegant one, but since the `currentmodule` sphinx directive is set last for tensorflow all the following doc generation takes place for tensorflow. Hence we need to make sure that the docs are under the respective tabs.

PR #2061, opened 2022-03-11
TF `try_gpu` string output fix

biswajitsahoo1111:
*Description of changes:*

This pull requrest only affects the string output that is generated after training. For example, the following old output

```
x tokens/sec on <tensorflow.python.eager.context._EagerDeviceContext object at 0x7ff8005deb00>
```
becomes
```
x tokens/sec on PhysicalDevice(name='/physical_device:GPU:0', device_type='GPU')
```
A similar change occurs for multiple GPUs and CPU. It doesn't affect the training process. At present, to the best of my knowledge, all TF code in D2L book uses only 1 GPU even when multiple GPUs are available. Using multiple GPUs in TF requires instantiating the model, loss function, etc. inside a strategy scope. Doing so might require changing the current TF code structure significantly from MXNet or PyTorch. But that's a discussion for another time. The current PR is not even remotely related to that.  

choice.

### DIFF
--- a/d2l/tensorflow.py
+++ b/d2l/tensorflow.py
@@ -14,12 +14,13 @@ import tarfile
 import time
 import zipfile
 from collections import defaultdict
+
+import numpy as np
 import pandas as pd
 import requests
 from IPython import display
 from matplotlib import pyplot as plt
 
-import numpy as np
 import tensorflow as tf
 
 d2l = sys.modules[__name__]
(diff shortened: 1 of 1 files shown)


biswajitsahoo1111:
It breaks the notebook in section 6.6. The problem to be addressed by this PR can be solved in another way (as done in section 6.6). So closing the PR.

PR #2074, opened 2022-03-18
fix the broken download link for movielens dataset

cheungdaven:
*Description of changes:*

choice.

### DIFF
--- a/chapter_recommender-systems/movielens.md
+++ b/chapter_recommender-systems/movielens.md
@@ -22,7 +22,7 @@ Then, we download the MovieLens 100k dataset and load the interactions as `DataF
 ```{.python .input  n=2}
 #@save
 d2l.DATA_HUB['ml-100k'] = (
-    'http://files.grouplens.org/datasets/movielens/ml-100k.zip',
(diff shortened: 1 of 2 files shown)


astonzhang:
Thanks. Can you run `d2lbook build lib` to save it into the d2l lib?

astonzhang:
@AnirudhDagar we'll need this fix for 0.17.5, just FYI

cheungdaven:
> Thanks. Can you run `d2lbook build lib` to save it into the d2l lib?

should not it be done automatically during compilation?

astonzhang:
> > Thanks. Can you run `d2lbook build lib` to save it into the d2l lib?
> 
> should not it be done automatically during compilation?

As part of changes to the source code in the repo, changes to d2l lib should be submitted by developers. In other words, developers run `d2lbook build lib` and include changes to d2l lib in the same PR.

AnirudhDagar:
> should not it be done automatically during compilation?
I was also thinking the same, some developers do not follow the d2lbook build lib process (or sometimes have an older version of d2lbook) and end up with different diffs.

PR #2121, opened 2022-05-10
Bug fixing for forward pass in PyTorch BERT

zxydi1992:
*Description of changes:*

The original forward with positional embedding use `.data`, which is detached from the computation graph. Instead, we should directly use `pos_embedding` in forward.

choice.


AnirudhDagar (approved):
Thanks @zxydi1992, nice catch! LGTM, since BERT uses learnable positional embeddings. I'll merge after testing this locally for subsequent sections.

astonzhang:
@zxydi1992 Thanks! Would you like to send another PR to replace your ID with your name in our acknowledgement:

zxydi1992:
> @zxydi1992 Thanks! Would you like to send another PR to replace your ID with your name in our acknowledgement:
> https://github.com/d2l-ai/d2l-en/blob/master/chapter_preface/index.md

I'll send the PR later. Feel pleasure for it, since I've learnt a lot through this awesome book.

Issue #2134, opened 2022-05-16
A mistake in seq2seq prediction implementation?


zhmou:
https://github.com/d2l-ai/d2l-en/blob/9e4fbb1e97f4e0b3919563073344368755fe205b/d2l/torch.py#L2996-L3030
**Bugs here:**
``` Python
    for _ in range(num_steps):
        Y, dec_state = net.decoder(dec_X, dec_state)
```
As you can see here, dec_state will update in every loop. But it not only affects the hidden state for rnn, but also the context vector for each step. (I don't know why Seq2SeqDecoder seems not implemented in [d2l/torch.py](https://github.com/d2l-ai/d2l-en/blob/9e4fbb1e97f4e0b3919563073344368755fe205b/d2l/torch.py)), In chapter 9.7.2, here wrote:
https://github.com/d2l-ai/d2l-en/blob/9e4fbb1e97f4e0b3919563073344368755fe205b/chapter_recurrent-modern/seq2seq.md?plain=1#L383-L411

![](https://raw.githubusercontent.com/d2l-ai/d2l-en/9e4fbb1e97f4e0b3919563073344368755fe205b/img/seq2seq-predict.svg)

And according to this graph, the correct approach should be to keep the context vector always constant at each time step.
(there is no problem with this method when training, because the target sentence is complete and context vector is already broadcast in every time step)

**My Solution(Not tested):**
Modify  Seq2SeqDecoder:
``` Python
class Seq2SeqDecoder(d2l.Decoder): 
     """The RNN decoder for sequence to sequence learning.""" 
     def __init__(self, vocab_size, embed_size, num_hiddens, num_layers, 
                  dropout=0): 
         super().__init__() 
         self.embedding = nn.Embedding(vocab_size, embed_size) 
         self.rnn = d2l.GRU(embed_size+num_hiddens, num_hiddens, 
                            num_layers, dropout) 
         self.dense = nn.LazyLinear(vocab_size) 
         self.apply(init_seq2seq) 
              
     def init_state(self, enc_outputs, *args): 
         return enc_outputs[1]  
     
     # Add a parameter here:
     def forward(self, X, enc_state, enc_final_layer_output_at_last_step): 
         # X shape: (batch_size, num_steps) 
         # embs shape: (num_steps, batch_size, embed_size) 
         embs = self.embedding(d2l.astype(d2l.transpose(X), d2l.int32)) 
         # context shape: (batch_size, num_hiddens)            
[...]

PR #2140, opened 2022-05-23
fix the tf bug in transformer.

cheungdaven:
*Description of changes:*

The flag training=True should be set, otherwise the model will not be trained.

The performance of tf is worse
choice.

### DIFF
--- a/chapter_attention-and-transformers/transformer.md
+++ b/chapter_attention-and-transformers/transformer.md
@@ -280,7 +280,7 @@ print('layer norm:', ln(X), '\nbatch norm:', bn(X))
 ln = tf.keras.layers.LayerNormalization()
 bn = tf.keras.layers.BatchNormalization()
 X = tf.constant([[1, 2], [2, 3]], dtype=tf.float32)
-print('layer norm:', ln(X), '\nbatch norm:', bn(X, training=True))
+print('layer norm:', ln(X), '\nbatch norm:', bn(X))
 ```
 
 Now we can implement the `AddNorm` class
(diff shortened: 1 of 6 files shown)


astonzhang:
@AnirudhDagar Since you had worked on this earlier, could you take a look first?

astonzhang:
> _Description of changes:_
> 
> By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

Could you add brief explanations of the fix? Thanks

cheungdaven:
@d2l-bot please rebuild

cheungdaven:
> @cheungdaven the `seq2seq` section still doesn't work. Maybe it needs a change in the learning rate.

@AnirudhDagar This pr is only for the transformer issue, I did not check the seq2seq issue.

AnirudhDagar:
I see, yes #2134 is definitely a bug that needs fixing but it is not the cause for the prediction issue in seq2seq.

Btw can you also add `training=False` here:
https://github.com/d2l-ai/d2l-en/blob/5aaf035bb728dbe731c435b36b3e7e195bdd78bc/chapter_recurrent-modern/seq2seq.md?plain=1#L284

https://github.com/d2l-ai/d2l-en/blob/5aaf035bb728dbe731c435b36b3e7e195bdd78bc/chapter_recurrent-modern/seq2seq.md?plain=1#L454

AnirudhDagar (approved):
Got it, thanks!

astonzhang:
Thanks for the fix!

PR #2152, opened 2022-06-10
[WIP] Editing xDL overview

astonzhang:
*Description of changes:*

By submitting this pull request, I confirm that you can use, modify,
copy, and redistribute this contribution, under the terms of your
choice.

PR #2154, opened 2022-06-10
Editing explainability overview

astonzhang:
*Description of changes:*

Following my edit https://github.com/d2l-ai/d2l-en/pull/2152

choice.


astonzhang:
After this PR is merged, consider the following improvement:

1. Use f-string in print(). See other sections for examples.
2. Consider removing BN from MLP to make it simpler.
3. Plot weights of 4 hidden layers + 1 output layer. Revise text accordingly.

astonzhang:
After this PR is merged, consider the following improvement:

Use f-string in print(). See other sections for examples.
Consider removing BN from MLP to make it simpler.
Plot weights of 4 hidden layers + 1 output layer. Revise text accordingly.

cheungdaven:
> After this PR is merged, consider the following improvement:
> 
>     * Use f-string in print(). See other sections for examples.
> 
>     * Consider removing BN from MLP to make it simpler.
> 
>     * Plot weights of 4 hidden layers + 1 output layer. Revise text accordingly.

There is an input layer and an output layer + three hidden layers. The input layer and output layer have different shapes so  we need more code to visualize them. Also the output layer shape is like 256*2, which is not that discernible in the figure.

chapter_explainability/explainability-overview.md
```diff
@@ -111,81 +112,88 @@ class HeartDiseaseMLP(d2l.Classifier): #@save
...
+Often, simple models such as linear regression, logistic regression, and decision trees are easy to interpret (or self-explanatory) but fall short in capturing the intricate patterns in data.
+Nonetheless,
+humans have been using easy-to-employ simple models to avoid cognitive overload. For example, emergency medical workers use the [rule of nines](https://en.wikipedia.org/wiki/Wallace_rule_of_nines) :numref:`fig_ruleof9s` to quickly assess a burn's percentage of the total skin; Recreational divers use a [diving planner](https://en.wikipedia.org/wiki/Recreational_Dive_Planner) to determine how long they can safely stay underwater at a given depth to avoid putting too much nitrogen in their bodies which can cause injury or even death.
```
cheungdaven, line 148: it seems that in this way, we are mixing the simple models in machine learning and other domains.

astonzhang: This just shows simple models exist for a reason, regardless of domains

PR #2199, opened 2022-07-14
Style

cheungdaven:
*Description of changes:*

remove ** because it will make the text in toc boldface.

choice.


astonzhang:
Actually it's slides syntax. So **x** should be [**x**]. Will fix

PR #2202, opened 2022-07-14
[TF] Implement TrainCallback on batch

adtygan:
*Description of changes:*

1.Updated class definition in https://github.com/d2l-ai/d2l-en/blob/master/chapter_appendix-tools-for-deep-learning/utils.md
2. However, when I ran d2lbook build lib, it removed imports in d2l/mxnet.py, d2l/torch.py and d2l/tensorflow.py. Therefore, I manually updated the class definition in d2l/tensorflow.py directly.

choice.

### DIFF
--- a/chapter_appendix-tools-for-deep-learning/utils.md
+++ b/chapter_appendix-tools-for-deep-learning/utils.md
@@ -73,17 +73,17 @@ def draw(self, x, y, label, every_n=1):
     if len(points) != every_n:
         return    
     mean = lambda x: sum(x) / len(x)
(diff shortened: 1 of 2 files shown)


adtygan:
The CI logs here - http://ci.d2l.ai/blue/organizations/jenkins/d2l-en/detail/PR-2202/1/pipeline/ show that the notebooks were re-run but still the plots don't seem to be updated. As seen here http://preview.d2l.ai/d2l-en/PR-2202/chapter_optimization/lr-scheduler.html , the Tensorflow plot is still smooth unlike other two. I'm not sure where the error is.

PR #2038, opened 2022-02-06
[WIP] switch to lazy api

mli:
*Description of changes:*

By submitting this pull request, I confirm that you can use, modify,
copy, and redistribute this contribution, under the terms of your
choice.

PR #2225, opened 2022-07-30
[PyTorch] Fix module types (LazyLinear -> Linear)

atgctg:
*Description of changes:*

[In 6.3 Parameter Initialization](http://d2l.ai/chapter_builders-guide/init-param.html), surprisingly `LazyLinear` modules have the type `<class 'torch.nn.modules.linear.Linear'>` during init, so the in-place updates in the PyTorch examples are skipped.

This PR fixes this by making the if statements check for the correct `nn.Linear` type.

Not applied:
```py
def init_constant(module):
    if type(module) == nn.LazyLinear: # this is False
        nn.init.constant_(module.weight, 1)
        nn.init.zeros_(module.bias)
net.apply(init_constant)
net[0].weight.data[0], net[0].bias.data[0]
```
```
(tensor([ 0.1594,  0.9592, -1.4913,  0.8780]), tensor(0.))
```

Applied:
```py
def init_constant(module):
    if type(module) == nn.Linear: # this is True
        nn.init.constant_(module.weight, 1)
        nn.init.zeros_(module.bias)
net.apply(init_constant)
net[0].weight.data[0], net[0].bias.data[0]
```
```
(tensor([1., 1., 1., 1.]), tensor(0.))
```

---

choice.

### DIFF
--- a/chapter_builders-guide/init-param.md
+++ b/chapter_builders-guide/init-param.md
@@ -89,7 +89,7 @@ net[0].weight.data()[0]
 ```{.python .input  n=6}
 %%tab pytorch
 def init_normal(module):
-    if type(module) == nn.LazyLinear:
+    if type(module) == nn.Linear:
         nn.init.normal_(module.weight, mean=0, std=0.01)
         nn.init.zeros_(module.bias)
 net.apply(init_normal)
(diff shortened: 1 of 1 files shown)


AnirudhDagar (approved):
Hi @atgctg thanks for the PR. Great catch with this bug, which was overlooked in #2038. It is actually not surprising for the lazy layers to be `torch.nn.modules.linear.Linear` type, since PyTorch devs designed the lazy modules intending to convert them to their non-lazy analogue after the first dry run (param init). See [docs](https://pytorch.org/docs/stable/generated/torch.nn.LazyLinear.html#torch.nn.LazyLinear.cls_to_become).

astonzhang:
@atgctg thanks. Could you send another PR to replace your github id with your name in our acknowledgement:

PR #2233, opened 2022-08-03
Update gru.md

azimjonn:
*Description of changes:*
Cleaned implementation of the forward method.

choice.


azimjonn:
`if H is None: H = torch.zeros(inputs.shape[1], self.num_hiddens)`

The above line should be changed to below:

`if H is None: H = torch.zeros(inputs.shape[1], self.num_hiddens, device=inputs.device)`

I am sorry if I did something wrong, this is my first pull request.

azimjonn:
Oh, that code was for all pytorch, mxnet and TensorFlow. I thought I changed only the pytorch code(.

PR #2240, opened 2022-08-08
Jacobian matrix

elizabethhsy:
*Description of changes:*

I'm new to linear algebra so I'm not sure if it's correct, but it seems like the Jacobian matrix for a function $f: \mathbb{R}^n \to \mathbb{R}^m$ should also have dimension $m \times n$. As the product $\mathbf{A} \mathbf{x}$ has such a mapping, its derivative would also have dimensions $m \times n$. Therefore $\nabla_{\mathbf{x}} \mathbf{A} \mathbf{x} = \mathbf{A}$, not $\mathbf{A}^\top$.

choice.

### DIFF
--- a/chapter_preliminaries/calculus.md
+++ b/chapter_preliminaries/calculus.md
@@ -312,7 +312,7 @@ by $\nabla f(\mathbf{x})$.
 The following rules come in handy 
 for differentiating multivariate functions:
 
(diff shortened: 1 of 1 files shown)


astonzhang:
Thanks. After checking we'll keep the original :)

PR #2249, opened 2022-08-15
Starting CNN modern merge

smolix:
*Description of changes:*

Porting the CNN modern changes in (RegNetX still pending). Figure cleanup first.

choice.


astonzhang:
Thanks! I'll polish the edit in new commits.

chapter_convolutional-modern/alexnet.md
```diff
@@ -195,18 +198,17 @@ By comparison, GPUs can consist of thousands of small processing elements (NIVID
...
 Hence, for the power budget of a CPU core that runs 4 times faster (a typical number),
 you can use 16 GPU cores at $\frac{1}{4}$ the speed,
 which yields $16 \times \frac{1}{4} = 4$ times the performance.
-Furthermore, GPU cores are much simpler
+Secondly, GPU cores are much simpler
 (in fact, for a long time they were not even *able*
 to execute general-purpose code),
-which makes them more energy efficient.
-Last, many operations in deep learning require high memory bandwidth.
+which makes them more energy efficient. For instance, they tend not to support speculative evaluation, it typically isn't possible to program each processing element individually, and the caches per core tend to be much smaller.
```
astonzhang, line 210: What's the logical flow among:
"they tend not to support speculative evaluation", "it typically isn't possible to program each processing element individually", and "the caches per core tend to be much smaller"?

chapter_convolutional-modern/resnet.md
```diff
@@ -384,63 +381,26 @@ with d2l.try_gpu():
...
-have `bot_channels//groups` groups.
-Similar to
-the residual block implementation in
-:numref:`subsec_residual-blks`,
-the residual connection
-is generalized
-with a $1 \times 1$ convolution (`conv4`),
-where setting `use_1x1conv=True, strides=2`
-halves the input height and width.
+Breaking up a convolution from $c$ to $b$ channels into one of $g$ groups of size $c/g$ generating $g$ outputs of size $b/g$ is called, quite fittingly, a *grouped convolution*. The computational cost is reduced from $O(c \cdot b)$ to $O(g \cdot (c/g) \cdot (b/g)) = O(c \cdot b / g)$, i.e. it is $g$ times faster. Even better, the number of parameters needed to generate the output is also reduced from a $c \times b$ matrix to $g$ smaller matrices of size $(c/g) \times (b/g)$, again a $g$ times reduction. In what follows we assume that both $b$ and $c$ are divisible by $g$.
```
astonzhang, line 394: According to Fig 3(b) of [the ResNeXt paper](https://openaccess.thecvf.com/content_cvpr_2017/papers/Xie_Aggregated_Residual_Transformations_CVPR_2017_paper.pdf), the input channel $c$ (256 in Fig 3(b)) remains the same (rather than being divided into $c/g$ channels) in every group. Thus,

"Breaking up a convolution from $c$ to $b$ channels into one of $g$ groups of size $c/g$ generating $g$ outputs of size $b/g$"
->
"Breaking up a convolution from $c$ to $b$ channels into one of $g$ groups of size $c$ generating $g$ outputs of size $b/g$"

Since this affects the subsequent arguments, could you please double check and revise if necessary? Please correct me if I am wrong. Thanks.

astonzhang (commented):
Thanks! Just two comments:

chapter_convolutional-modern/alexnet.md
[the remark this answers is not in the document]
smolix: Well, these are three separate things. As in, apples, bananas and citrons. 

Speculative evaluation: where a code path is executed even before the logical true/false in a conditional evaluation is known. This allows for more efficient computation (processor doesn't stall) but also for potentially wasted results and the need to undo changes. 

Program each element individually: within a warp you have the same code executed on every stream processor. For modern CPUs each core can do whatever it wants. 

GPUs don't have a lot of cache but they have super fast RAM (HBM2 and better on the high end, GDDR6 on the low end). Even the RTX3090Ti has 1TB/s memory bandwidth. Graviton / Xeon / Epyc have up to 200GB/s, desktop processors around 100GB/s. But then even a puny desktop CPU has 16-32GB L3 cache.

chapter_convolutional-modern/resnet.md
smolix: No, your interpretation isn't correct. The point is that you have an initial bottleneck to go from c (channel) to b (bottleneck) dimensions. After that, each grouped convolution (there are g groups) acts only on a small slice of b/g inputs and produces b/g outputs. Otherwise there wouldn't be any benefit in using ResNeXt.

astonzhang: Agreed. So $c$ and $b$ in text should be $c_{in}$ and $c_{out}$ in general, not to be confused with $c$ and $b$ in Fig.8.6.5. Actually for the grouped convolution in Fig.8.6.5, $c_{in} = c_{out} = b$.

PR #2260, opened 2022-08-20
review number of flops operation per sec

Excelsior7:
The multiplication of two matrices with an identical form of (256,256) implies 256^3 = 16777216 multiplications of real numbers which is approximately 0.02*10^9. So unless I have omitted something in my reasoning, the 2 must be replaced by 0.02 while keeping the rounding to simplify.

### DIFF
--- a/chapter_optimization/minibatch-sgd.md
+++ b/chapter_optimization/minibatch-sgd.md
@@ -207,7 +207,7 @@ A.assign(tf.tensordot(B, C, axes=1))
 timer.stop()
 
 # Multiply and add count as separate operations (fused in practice)
-gigaflops = [2/i for i in timer.times]
+gigaflops = [0.02/i for i in timer.times]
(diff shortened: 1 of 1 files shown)


cheungdaven:
@Excelsior7 @astonzhang Since Multiply and add count as separate operations , will 0.02 * 2 make more sense?

Excelsior7:
> @Excelsior7 @astonzhang Since Multiply and add count as separate operations , will 0.02 * 2 make more sense?

Indeed, to follow the logic of the comment before the code snippet, it makes more sense. I modify this and create a new commit.

astonzhang:
I made some further edits: https://github.com/d2l-ai/d2l-en/commit/3b40fcefddb91d5989df3b70e96fa5be155a77b6
Just let me know if that looks good to you.

Excelsior7:
Introducing how this quantity was calculated is far from being too much, on the contrary. And I agree with you that 0.03 is closer to reality. Great!

PR #2271, opened 2022-08-26
Increase number of iterations for MLE code

JojiJoseph:
*Description of changes:*
Increased the number of iterations for MLE code, so that in the output the predicted theta will be same as actual theta. Also increase number of tails.

choice.

### DIFF
--- a/chapter_appendix-mathematics-for-deep-learning/maximum-likelihood.md
+++ b/chapter_appendix-mathematics-for-deep-learning/maximum-likelihood.md
@@ -135,8 +135,8 @@ theta = np.array(0.5)
 theta.attach_grad()
 
 # Perform gradient descent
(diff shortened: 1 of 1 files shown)


astonzhang:
Thanks. It's not necessary to make the estimate *exactly* equal to the ground-truth. It's more ideal to have a close estimate while reducing training time in our demo :)

JojiJoseph:
> It's more ideal to have a close estimate while reducing training time in our demo :)

Previously, it wasn't even a close estimate. Estimated: **0.5017**, GT: **0.9970**.

Also, `lr` was `1e-11`. If we increase `lr` to `1e-9` we could get a close approximation in `100` iterations.

Anyway, if it is not necessary to update, I can close this pull request.

PR #2274, opened 2022-08-28
Update kaggle-dog.md

Sere1nz:
In line 495, softmax should be done on dim=1. Ouput shape of net is (128,120). dim = 0 is the dimension of batchsize. Softmax should be done over dim=1 which is 120 classes.

*Description of changes:*

choice.


AnirudhDagar (approved):
Thanks @Sere1nz for the fix, that is indeed correct.

AnirudhDagar:
@Sere1nz Could you please send another PR to replace your id with your name in our [acknowledgement](https://github.com/d2l-ai/d2l-en/blob/master/chapter_preface/index.md)?

PR #2278, opened 2022-08-29
BUG: Fix broken preview path for release

AnirudhDagar:
The logic for release preview is updated.

choice.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -52,7 +52,7 @@ stage("Build and Publish") {
 
       sh label:"Build HTML", script:"""set -ex
       conda activate ${ENV_NAME}
-      ./static/build_html.sh ${env.BRANCH_NAME} ${JOB_NAME}
+      ./static/build_html.sh ${env.BRANCH_NAME} ${JOB_NAME} ${REPO_NAME}
       """
 
       sh label:"Build PDF", script:"""set -ex
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -52,7 +52,7 @@ stage("Build and Publish") {
 
       sh label:"Build HTML", script:"""set -ex
       conda activate ${ENV_NAME}
-      ./static/build_html.sh ${env.BRANCH_NAME} ${JOB_NAME} ${REPO_NAME}
+      ./static/build_html.sh ${env.BRANCH_NAME} ${JOB_NAME}
       """
 
       sh label:"Build PDF", script:"""set -ex
--- a/static/build_html.sh
+++ b/static/build_html.sh
@@ -5,16 +5,16 @@ set -e
 # Read arguments
 BRANCH_NAME=$1
 JOB_NAME=$2
+REPO_NAME=$3
 
 STABLE_BASE_PATH="https://d2l.ai"
 PREVIEW_BASE_PATH="http://preview.d2l.ai/$JOB_NAME"
-PREVIEW_BASE_PATH_RELEASE="${PREVIEW_BASE_PATH/release/master}"
+alternate_base="http://preview.d2l.ai/$REPO_NAME/master"
 
 # Generate Headers
 if [[ "$BRANCH_NAME" == "release" ]]; then
     echo "Use release headers"
     alternate_text="Preview Version"
-    alternate_base=$PREVIEW_BASE_PATH_RELEASE
     current_base=$STABLE_BASE_PATH
 else
     echo "Use ${JOB_NAME} headers"
--- a/static/build_html.sh
+++ b/static/build_html.sh
@@ -5,17 +5,15 @@ set -e
 # Read arguments
 BRANCH_NAME=$1
 JOB_NAME=$2
-REPO_NAME=$3
 
 STABLE_BASE_PATH="https://d2l.ai"
 PREVIEW_BASE_PATH="http://preview.d2l.ai/$JOB_NAME"
-alternate_base="http://preview.d2l.ai/$REPO_NAME/master"
 
 # Generate Headers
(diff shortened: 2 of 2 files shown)


static/build_html.sh
```diff
@@ -8,12 +8,13 @@ JOB_NAME=$2
 
 STABLE_BASE_PATH="https://d2l.ai"
 PREVIEW_BASE_PATH="http://preview.d2l.ai/$JOB_NAME"
+PREVIEW_BASE_PATH_RELEASE="${PREVIEW_BASE_PATH/release/master}"
```
astonzhang, line 11: What is release/master?

astonzhang: should be http://preview.d2l.ai/d2l-en/master, where d2l-en is extracted from JOB_NAME?

AnirudhDagar: This is bash syntax to replace release with master. Notice the curly braces.

astonzhang: Maybe we only need REPO_NAME? https://github.com/d2l-ai/d2l-en/blob/master/Jenkinsfile#L3

AnirudhDagar: There are multiple ways to implement this. But in any case, we cannot avoid JOB_NAME because it carries the PR number information which is required when PR previews are build and pdfs are correctly linked. So instead of passing one more argument, I just used the information from JOB_NAME.

Feel free to suggest an edit.

astonzhang: Have you considered r0.17 branch and classic branch?

astonzhang: The variable name `PREVIEW_BASE_PATH_RELEASE` is not readable. Why not simply:

```
alternate_base="${PREVIEW_BASE_PATH/release/master}"
```

astonzhang: Otherwise LGTM

AnirudhDagar: > Have you considered r0.17 branch and classic branch?

Good catch, this logic is only for the release branch. It will not work on classic branch. Are we hosting r0.17 somewhere stably? r0.17 seems like just another branch now, right?

d2l.ai hosts the release. (STABLE)
classic.d2l.ai hosts the classic. (STABLE)
All other branches are like preview branches. (PREVIEW)

I can add a specific logic for classic like L14 and also give it separate variables.

AnirudhDagar: But in any case this is differnt for classic branch because we will never merge those. When sending a PR for classic branch i can add it there if we want this feature to be backported to the classic branch.

astonzhang: Since r0.17 is in maintenance mode, let's only implement this change on actively-developed master/release

AnirudhDagar: Ok in that case this logic should be fine.

> The variable name PREVIEW_BASE_PATH_RELEASE is not readable. Why not simply:

I was just trying to be consistent with the naming. alternate_base was supposed to be calculated inside the if else but I'm fine with your suggested change if that looks better to you.

AnirudhDagar:
I've made the code more readable, addressing the comments. Please review again @astonzhang thanks! 🙌🏼

Jenkinsfile
```diff
@@ -52,7 +52,7 @@ stage("Build and Publish") {
 
       sh label:"Build HTML", script:"""set -ex
       conda activate ${ENV_NAME}
-      ./static/build_html.sh ${env.BRANCH_NAME} ${JOB_NAME}
+      ./static/build_html.sh ${env.BRANCH_NAME} ${JOB_NAME} ${REPO_NAME}
```
astonzhang, line 55:
```suggestion
      ./static/build_html.sh ${env.BRANCH_NAME} ${JOB_NAME}
```

static/build_html.sh
```diff
@@ -5,16 +5,16 @@ set -e
 # Read arguments
 BRANCH_NAME=$1
 JOB_NAME=$2
+REPO_NAME=$3
```
astonzhang, line 8:
```suggestion
```

static/build_html.sh
```diff
@@ -5,16 +5,16 @@ set -e
...
 
 STABLE_BASE_PATH="https://d2l.ai"
 PREVIEW_BASE_PATH="http://preview.d2l.ai/$JOB_NAME"
-PREVIEW_BASE_PATH_RELEASE="${PREVIEW_BASE_PATH/release/master}"
+alternate_base="http://preview.d2l.ai/$REPO_NAME/master"
```
astonzhang, line 12:
```suggestion
```

static/build_html.sh
```diff
@@ -5,16 +5,16 @@ set -e
...
 
 # Generate Headers
 if [[ "$BRANCH_NAME" == "release" ]]; then
     echo "Use release headers"
     alternate_text="Preview Version"
```
astonzhang, line 17:
```suggestion
    alternate_text="Preview Version"
    alternate_base="${PREVIEW_BASE_PATH/release/master}"
```

astonzhang: This is what I meant. Otherwise your line 12 `alternate_base="http://preview.d2l.ai/$REPO_NAME/master"` will be overwritten by line 22 `alternate_base=$STABLE_BASE_PATH` in that branch: this may be confusing

AnirudhDagar: Got it, thanks for the changes.

AnirudhDagar: It is still overwritten. No?

AnirudhDagar: Ok sorry, i don't know what happened. Seems like batched applied suggestions did not work correctly.

PR #2279, opened 2022-08-29
Depend on matplotlib-inline instead of ipython to avoid colab warning

### LINKED ISSUE #2250: ModuleNotFoundError when running the official pytorch colab notebook
image

I can replicate the error at multiple official pytorch colab notebooks, e.g. 

https://colab.research.google.com/github/d2l-ai/d2l-pytorch-colab/blob/master/chapter_linear-classification/image-classification-dataset.ipynb#scrollTo=ee445cce

AnirudhDagar:
colab warns when using ipython>1.0.0, hence avoid depending on newer version of ipython. Instead directly depend on matplotlib-inline to fix #2250 

image (5)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@ from setuptools import setup, find_packages
 import d2l
 
 requirements = [
-    'ipython>=7.23',
+    'matplotlib-inline',
     'jupyter',
     'numpy',
     'matplotlib',
--- a/setup.py
+++ b/setup.py
@@ -2,10 +2,10 @@ from setuptools import setup, find_packages
 import d2l
 
 requirements = [
-    'matplotlib-inline',
     'jupyter',
     'numpy',
     'matplotlib',
+    'matplotlib-inline',
     'requests',
     'pandas',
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -2,7 +2,7 @@
 import d2l
 
 requirements = [
-    'ipython>=7.23',
+    'matplotlib-inline',
```
astonzhang, line 5:
```suggestion
```

setup.py
```diff
@@ -2,7 +2,7 @@
...
     'jupyter',
     'numpy',
     'matplotlib',
```
astonzhang, line 8:
```suggestion
    'matplotlib',
    'matplotlib-inline',
```

astonzhang (changes requested):

astonzhang:
Let's group matplotlib-related lib together. Otherwise LGTM

setup.py
AnirudhDagar: Thanks, made the change.

PR #2034, opened 2022-02-01
Fix issues of weight-decay

cheungdaven:
1. In MXNet, the concise implementation of using weight decay generates a plot that looks like no weight decay is used.
weight decay parameter is added.

2. In TensorFlow, the concise implementation and from-scratch implementation plots (using weight decay) do not align with each other well.
The hyper-parameter is tuned.

*Description of changes:*

By submitting this pull request, I confirm that you can use, modify,
copy, and redistribute this contribution, under the terms of your
choice.

PR #912, opened 2020-05-31
pytorch: weight decay

AnirudhDagar:
*Description of changes:*

By submitting this pull request, I confirm that you can use, modify,
copy, and redistribute this contribution, under the terms of your
choice.

PR #2284, opened 2022-08-30
Wrong statement regarding weight decay in pytorch

finale80:
The two sentences 
*[...] By default, PyTorch decays both weights and biases simultaneously. Here, we only set weight_decay for the weight, so the bias parameter b will not decay [...]*
contradict each other. More specifically, I believe the 2nd one is wrong.

As from PyTorch implementation https://github.com/pytorch/pytorch/blob/acd6ca8cfa9537284928fb5d36834d1e5ae1e6f3/torch/optim/sgd.py#L120
all parameters receive the same weight decay.
So the first sentence is correct.

In the example of this chapter, to avoid decaying the bias, one would need to use the following
```
...
   def loss(self, y_hat, y):
        return super().loss(y_hat, y) + self.weight_decay * 0.5 * (self.net.weight**2).sum()

    def configure_optimizers(self):
        return torch.optim.SGD(self.parameters(), lr=self.lr, weight_decay=0)
...
```

I attach here a self-contained example based on this chapter
(note: I used .txt to bypass file extension checking).

The code compares 3 versions of the model
1. the default implementation (i.e., the one suggested by the book which decays all parameters)
2. a "manual" implementation of the default behaviour (to show that the default is indeed decaying all parameters)
3. the option mentioned above (which does not decay the bias).

Overall, the simplest way to fix the problem is to drop the 2nd sentence as proposed.

Alternatively, one could add the comparison as from the code attached. But
1. I need to be extended to Tensorflow and MXNet
6. I think the right way to skip the decay would be by specifying the parameter differently (I believe one could group parameters with different options). But this would be too much details given the aim of the chapter

*Description of changes:*

choice.

[chapter3.7_weight-decay.py.txt](https://github.com/d2l-ai/d2l-en/files/9454152/chapter3.7_weight-decay.py.txt)

### DIFF
--- a/chapter_linear-regression/weight-decay.md
+++ b/chapter_linear-regression/weight-decay.md
@@ -398,13 +398,10 @@ class WeightDecay(d2l.LinearRegression):
 
    def configure_optimizers(self):
         """Defined in :numref:`sec_linear_concise`"""
-        return torch.optim.SGD(
-            [
+        return torch.optim.SGD([
                 {"params": self.net.weight, "weight_decay": self.wd},
-                {"params": self.net.bias},
-            ],
-            lr=self.lr,
-        )
+                {"params": self.net.bias}],
+            lr=self.lr)
 ```
 
 ```{.python .input  n=13}
--- a/chapter_linear-regression/weight-decay.md
+++ b/chapter_linear-regression/weight-decay.md
@@ -397,7 +397,6 @@ class WeightDecay(d2l.LinearRegression):
         self.wd = wd
 
     def configure_optimizers(self):
-        """Defined in :numref:`sec_linear_concise`"""
         return torch.optim.SGD([
             {'params': self.net.weight, 'weight_decay': self.wd},
             {'params': self.net.bias}], lr=self.lr)
--- a/chapter_linear-regression/weight-decay.md
+++ b/chapter_linear-regression/weight-decay.md
@@ -359,7 +359,12 @@ In the following code, we specify
 the weight decay hyperparameter directly
 through `weight_decay` when instantiating our optimizer.
 By default, PyTorch decays both
-weights and biases simultaneously.
+weights and biases simultaneously, but
+we can configure the optimizer to handle different parameters
+according to different policies.
+Here, we only set `weight_decay` for
+the weights (the `net.weight` parameters), hence the 
+bias (the `net.b` parameter) will not decay.
 :end_tab:
 
 :begin_tab:`tensorflow`
--- a/chapter_linear-regression/weight-decay.md
+++ b/chapter_linear-regression/weight-decay.md
@@ -453,10 +453,10 @@ to all layers of a deep network.
 
 ## Summary
 
-* Regularization is a common method for dealing with overfitting. Classical regularization techniques add a penalty term to the loss function (when training) to reduce the complexity of the learned model.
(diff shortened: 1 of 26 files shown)


AnirudhDagar (commented):
Thanks @finale80 for your PR!

This is a good catch and makes sense now. I remember adding this statement earlier because it was actually relevant at the time of writing the code for pytorch version (see #912) in the old d2l api. We used to only decay the weight params and not the bias.

So we had something like the following to handle weight decay separately for pytorch:

```python
      return torch.optim.SGD([
          {"params":self.net.weight,'weight_decay': self.wd},
          {"params":self.net.bias}], lr=self.lr)
```

I see that the above code was refactored and changed in #2034 and the text related to the code change was probably missed.

@cheungdaven after PR #2034 there seems to be an inconsistency between pytorch and mxnet. Specifically mxnet doesn't decay bias but pytorch does. I'm curious if there was any specific reason to remove [these L394-396](https://github.com/d2l-ai/d2l-en/pull/2034/files#diff-8be31f544cfaa6008c11c863dba743c9a4abb4a0436abb2763810b701d41fed0L394-L396)?

cheungdaven:
It makes sense to me. feel free to change it back @AnirudhDagar

AnirudhDagar:
Thanks @cheungdaven! @finale80 could you please add the params grouping back in pytorch code like we had before instead of editing the text?

finale80:
@AnirudhDagar done!

More specifically
* I've updated the code
* I re-integrated the original statement in the text (removed in my PR) and added minor modification to the text (to make it even more clear).

I also double checked the code and indeed everything is consistent now.
I attach my updated script.](url)
[chapter3.7_weight-decay.py.txt](https://github.com/d2l-ai/d2l-en/files/9540129/chapter3.7_weight-decay.py.txt)

chapter_linear-regression/weight-decay.md
```diff
@@ -392,10 +395,16 @@ class WeightDecay(d2l.LinearRegression):
...
-        return torch.optim.SGD(self.net.parameters(), 
-                               lr=self.lr, weight_decay=self.wd)
+        """Defined in :numref:`sec_linear_concise`"""
+        return torch.optim.SGD(
+            [
+                {"params": self.net.weight, "weight_decay": self.wd},
+                {"params": self.net.bias},
+            ],
+            lr=self.lr,
+        )
```
AnirudhDagar, line 407:
```suggestion
        return torch.optim.SGD([
            {'params': self.net.weight, 'weight_decay': self.wd},
            {'params': self.net.bias}], lr=self.lr)
```

AnirudhDagar (changes requested):
@finale80 just a minor style comment below (the idea is to minimize usage of lines wherever possible), the rest looks good. Thanks, will merge this after the change.

chapter_linear-regression/weight-decay.md
```diff
@@ -392,10 +395,12 @@ class WeightDecay(d2l.LinearRegression):
         super().__init__(lr)
         self.save_hyperparameters()
         self.wd = wd
-    
+
     def configure_optimizers(self):
-        return torch.optim.SGD(self.net.parameters(), 
-                               lr=self.lr, weight_decay=self.wd)
+        """Defined in :numref:`sec_linear_concise`"""
```
AnirudhDagar, line 400: Also no need to add this line.
```suggestion
```

AnirudhDagar (approved):

PR #2287, opened 2022-09-01
Add encoding with open file

nantekoto:
*Description of changes:*
assign utf-8 to open() to avoid encoding error

choice.

### DIFF
--- a/d2l/torch.py
+++ b/d2l/torch.py
@@ -2897,7 +2897,7 @@ def read_data_nmt():
 
     Defined in :numref:`sec_utils`"""
     data_dir = d2l.download_extract('fra-eng')
-    with open(os.path.join(data_dir, 'fra.txt'), 'r') as f:
+    with open(os.path.join(data_dir, 'fra.txt'), 'r', encoding='utf-8') as f:
         return f.read()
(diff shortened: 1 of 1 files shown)


AnirudhDagar (commented):
@nantekoto can you please share what exactly your environment is so that we can reproduce the mentioned encoding issue? I'm guessing you are on a windows machine?

I don't see this error and will need more details to reproduce it.

nantekoto:
> @nantekoto can you please share what exactly your environment is so that we can reproduce the mentioned encoding issue? I'm guessing you are on a windows machine?
> 
> I don't see this error and will need more details to reproduce it.

yes, on windows machine in Chinese the default decoding is gbk, which will be assignd to open() as well and result in an error as follow:

UnicodeDecodeError: 'gbk' codec can't decode byte 0xaf in position 33: illegal multibyte sequence

astonzhang:
Thanks. I think it's fixed in http://zh-v2.d2l.ai.s3-website-us-west-2.amazonaws.com/chapter_recurrent-modern/machine-translation-and-dataset.html so let's apply the change here, too.

PR #2305, opened 2022-09-14
JAX: Add 6.5 Custom Layers

AnirudhDagar:
*Description of changes:*

choice.

### DIFF
--- a/chapter_builders-guide/custom-layer.md
+++ b/chapter_builders-guide/custom-layer.md
@@ -28,7 +28,7 @@ from the base layer class and implement the forward propagation function.
 
 ```{.python .input}
 %load_ext d2lbook.tab
-tab.interact_select(['mxnet', 'pytorch', 'tensorflow'])
+tab.interact_select(['mxnet', 'pytorch', 'tensorflow', 'jax'])
 ```
 
 ```{.python .input}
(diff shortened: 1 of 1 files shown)


chapter_builders-guide/custom-layer.md
```diff
@@ -102,12 +114,23 @@ net = nn.Sequential(nn.LazyLinear(128), CenteredLayer())
...
+```
+
 As an extra sanity check, we can send random data
 through the network and check that the mean is in fact 0.
 Because we are dealing with floating point numbers,
 we may still see a very small nonzero number
 due to quantization.
 
+:begin_tab:`jax`
+Here we utilize the method `init_with_output` which returns both the output of
```
astonzhang, line 129: the `init_with_output` method

chapter_builders-guide/custom-layer.md
```diff
@@ -233,6 +287,12 @@ linear(torch.rand(2, 5))
 dense(tf.random.uniform((2, 5)))
 ```
 
+```{.python .input}
+%%tab jax
+dense.apply(params, jax.random.uniform(jax.random.PRNGKey(d2l.get_seed()),
+                                        (2, 5)))
```
astonzhang, line 293: Plz align this line of code (left shift)

PR #2307, opened 2022-09-15
JAX: Add 6.4 Lazy Initialization

AnirudhDagar:
This PR adds the section on lazy initialization for JAX. I've also taken this opportunity to refactor the method `apply_init` and move it to a more suitable class i.e `d2l.Module` instead of `d2l.Trainer`. Even though this method is a must for any Flax/Jax implementation, since the user has to handle the param inits, I've postponed it's implementation to a more suitable section i.e Lazy Init (similar to how it is done in PyTorch). But please note that we still make use of this function even before section 6.4.

This is like how we define `save_hyperparameters` only in section [20.7. Utility Functions and Classes](https://d2l.ai/chapter_appendix-tools-for-deep-learning/utils.html), but we start using it as soon as section [3.2. Object-Oriented Design for Implementation](https://d2l.ai/chapter_linear-regression/oo-design.html).

Also I think it would be best to rerun all the JAX notebooks with this change just to make sure that this change in API doesn't break anything. I'll wait to rerun CI (with cache cleaned) on this, after merging the other chapter 6 PRs that are open atm.

### DIFF
--- a/chapter_builders-guide/lazy-init.md
+++ b/chapter_builders-guide/lazy-init.md
@@ -95,8 +95,8 @@ We confirm by attempting to access the parameters below.
 
 :begin_tab:`jax`
 As mentioned previously, parameters and the network definition are decoupled
-in Jax and Flax, the user handles both manually. Flax models are stateless
-hence there is no parameters attribute.
+in Jax and Flax, and the user handles both manually. Flax models are stateless
+hence there is no `parameters` attribute.
 :end_tab:
 
 ```{.python .input}
@@ -204,7 +204,7 @@ It will be used later when default random initializations are not desired.
 :end_tab:
 
 :begin_tab:`jax`
-Parameter Initialization in Flax is always done manually and handled by the
+Parameter initialization in Flax is always done manually and handled by the
 user. The following method sets a PRNG Key and passes in a dummy input to
 initialize the model parameters and return the parameters.
 We have been using it under the hood in the previous sections as well.
--- a/chapter_builders-guide/lazy-init.md
+++ b/chapter_builders-guide/lazy-init.md
@@ -94,7 +94,7 @@ We confirm by attempting to access the parameters below.
 :end_tab:
 
 :begin_tab:`jax`
-As mentioned previously, parameters and the network definition are decoupled
+As mentioned in :numref:`subsec_param-access`, parameters and the network definition are decoupled
 in Jax and Flax, and the user handles both manually. Flax models are stateless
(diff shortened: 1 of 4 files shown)


chapter_builders-guide/lazy-init.md
```diff
@@ -74,11 +74,30 @@ net = tf.keras.models.Sequential([
...
 the dimensions of the input layer's weights
 because the input dimension remains unknown.
+
+:begin_tab:`mxnet, pytorch, tensorflow`
 Consequently the framework has not yet initialized any parameters.
 We confirm by attempting to access the parameters below.
+:end_tab:
+
+:begin_tab:`jax`
+As mentioned previously, parameters and the network definition are decoupled
```
astonzhang, line 97: "previously" is vague. Be specific where it was mentioned by using :numref:

chapter_builders-guide/lazy-init.md
```diff
@@ -74,11 +74,30 @@ net = tf.keras.models.Sequential([
...
+in Jax and Flax, the user handles both manually. Flax models are stateless
```
astonzhang, line 98: and the user handles both manually

chapter_builders-guide/lazy-init.md
```diff
@@ -74,11 +74,30 @@ net = tf.keras.models.Sequential([
...
+hence there is no parameters attribute.
```
astonzhang, line 99:
```
`parameters` attribute
```

chapter_builders-guide/lazy-init.md
```diff
@@ -178,6 +203,13 @@ and subsequently initializes the parameters.
 It will be used later when default random initializations are not desired.
 :end_tab:
 
+:begin_tab:`jax`
+Parameter Initialization in Flax is always done manually and handled by the
```
astonzhang, line 207: Initialization -> initialization

astonzhang (commented):
Next time you may consider grouping this PR with its dependency (if it's not that big). So it can be evaluated from scratch.

AnirudhDagar:
> Next time you may consider grouping this PR with its dependency (if it's not that big). So it can be evaluated from scratch.

Everything depends on this PR. Also this change was only understood when i started working on Ch6, hence i did not make the change originally. So it was not possible for me to add this earlier.

chapter_builders-guide/lazy-init.md
AnirudhDagar: Done!

PR #2313, opened 2022-09-26
Fix the numbering of the "burning questions"

vBarbaros:
Correct the wrong numbering in which all the questions were enumerated with 1.

### DIFF
--- a/chapter_linear-classification/generalization-classification.md
+++ b/chapter_linear-classification/generalization-classification.md
@@ -30,8 +30,8 @@ on random guessing whenever we encounter new examples.
 A number of burning questions demand immediate attention:
 1. How many test examples do we need to precisely estimate
    the accuracy of our classifiers on the underlying population?
-1. What happens if we keep evaluating models on the same test repeatedly?
-1. Why should we expect that fitting our linear models to the training set
+2. What happens if we keep evaluating models on the same test repeatedly?
+3. Why should we expect that fitting our linear models to the training set
    should fare any better than our naive memorization scheme?
 
 


AnirudhDagar (commented):
@vBarbaros, we have this in tonnes of notebooks. I don't think this is a necessary change. It should still render fine.

vBarbaros:
Hi @AnirudhDagar ,
Are you saying that since this error/typo is all over the place, it doesn't need to be corrected?
Br,
Victor

chapter_linear-classification/generalization-classification.md
```diff
@@ -30,8 +30,8 @@ on random guessing whenever we encounter new examples.
 A number of burning questions demand immediate attention:
 1. How many test examples do we need to precisely estimate
```
astonzhang, line 31:
```suggestion

1. How many test examples do we need to precisely estimate
```

chapter_linear-classification/generalization-classification.md
```diff
@@ -30,8 +30,8 @@ on random guessing whenever we encounter new examples.
...
    the accuracy of our classifiers on the underlying population?
-1. What happens if we keep evaluating models on the same test repeatedly?
-1. Why should we expect that fitting our linear models to the training set
+2. What happens if we keep evaluating models on the same test repeatedly?
```
astonzhang, line 33:
```suggestion
1. What happens if we keep evaluating models on the same test repeatedly?
```

chapter_linear-classification/generalization-classification.md
```diff
@@ -30,8 +30,8 @@ on random guessing whenever we encounter new examples.
...
+3. Why should we expect that fitting our linear models to the training set
```
astonzhang, line 34:
```suggestion
1. Why should we expect that fitting our linear models to the training set
```

PR #2314, opened 2022-09-30
Explicitly initialize RNN/LSTM/GRU states

gvtulder:
*Description of changes:*

The current from-scratch implementations of the RNN, LSTM and GRU do some complicated juggling with the initial states of the recurrent module.

Example for the current LSTM:
https://github.com/d2l-ai/d2l-en/blob/f4432f33be5ebecf709bec97cbf772e986b8da7b/chapter_recurrent-modern/lstm.md?plain=1#L286-L304

While this is clever and avoids a little bit of code, it has a few downsides:

* The initial state is initialized *inside* the for loop, which requires an `if H is None` block that runs at every time step, but only does something in the first step of the RNN. From a good programming perspective, this `if ` should be placed outside the loop.
* This lazy initialization obscures the fact that we need to initialize the state of the RNN to something before the first time step. I think this makes the model more difficult to understand than necessary. (For example, I had a discussion with a student where this initialization led to some confusion.)

With this pull request, I'd like to suggest some changes to initialize the states explicitly before the for loop. This requires one or two extra lines of code, but I think it would look cleaner and might make the models easier to follow.

Note that I added a `self.num_hiddens` to the RNN/LSTM/GRU classes. This is isn't absolutely necessary, you could also get this from the shape of the weights, but I think giving it a name might help.

-----
By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


chapter_recurrent-neural-networks/rnn-scratch.md
```diff
@@ -57,6 +57,7 @@ class RNNScratch(d2l.Module):  #@save
     def __init__(self, num_inputs, num_hiddens, sigma=0.01):
         super().__init__()
         self.save_hyperparameters()
+        self.num_hiddens = num_hiddens
```
astonzhang, line 60:
```suggestion
```

chapter_recurrent-modern/gru.md
```diff
@@ -175,6 +175,7 @@ class GRUScratch(d2l.Module):
     def __init__(self, num_inputs, num_hiddens, sigma=0.01):
         super().__init__()
         self.save_hyperparameters()
+        self.num_hiddens = num_hiddens
```
astonzhang, line 178:
```suggestion
```

chapter_recurrent-modern/lstm.md
```diff
@@ -254,6 +254,7 @@ class LSTMScratch(d2l.Module):  #@save
     def __init__(self, num_inputs, num_hiddens, sigma=0.01):
         super().__init__()
         self.save_hyperparameters()
+        self.num_hiddens = num_hiddens
```
astonzhang, line 257:
```suggestion
```

astonzhang (approved):
Thanks. I removed `self.num_hiddens = num_hiddens` because it's automatically done by self.save_hyperparameters()

astonzhang:
@gvtulder We've added your name in our acknowledgement https://github.com/d2l-ai/d2l-en/blob/master/chapter_preface/index.md 
Feel free to send another PR if there's any mistake in the name :)

PR #2315, opened 2022-10-01
JAX: Add dropout.md

AnirudhDagar:
*Description of changes:*
Note that dropout in flax requires passing rngs with a specific 'dropout' key. This is not required otherwise and is set to None by default. This PR handles that. In the future, we can refactor this further to make use of `make_rng` to make sure every dropout layer has a separate generation.

I also save a utility lambda function which can be used in later sections called `get_key`.

choice.

### DIFF
--- a/chapter_multilayer-perceptrons/dropout.md
+++ b/chapter_multilayer-perceptrons/dropout.md
@@ -201,9 +201,7 @@ from flax import linen as nn
 import jax
 from jax import numpy as jnp
 
-get_key = lambda: jax.random.PRNGKey(d2l.get_seed())  #@save
-
-def dropout_layer(X, dropout, key=get_key()):
+def dropout_layer(X, dropout, key=d2l.get_key()):
     assert 0 <= dropout <= 1
     if dropout == 1: return jnp.zeros_like(X)
     mask = jax.random.uniform(key, X.shape) > dropout
@@ -427,7 +425,7 @@ class DropoutMLP(d2l.Classifier):
         return nn.Dense(self.num_outputs)(x)
 
 @d2l.add_to_class(DropoutMLP)
-def loss(self, params, X, Y, averaged=True, rngs={'dropout': get_key()}):
+def loss(self, params, X, Y, averaged=True, rngs={'dropout': d2l.get_key()}):
     return super(DropoutMLP, self).loss(params, X, Y, averaged, rngs)
 ```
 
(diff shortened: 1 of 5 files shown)


chapter_multilayer-perceptrons/dropout.md
```diff
@@ -358,12 +407,39 @@ class DropoutMLP(d2l.Classifier):
...
 Next, we [**train the model**].
 
 ```{.python .input}
 %%tab all
 model = DropoutMLP(**hparams)
-trainer.fit(model, data)
+if tab.selected('pytorch', 'mxnet', 'tensorflow'):
+    trainer.fit(model, data)
+if tab.selected('jax'):
+    trainer.fit(model, data, key={'params': get_key(), 'dropout': get_key()})
```
astonzhang, line 442: Can we hide this `key` argument in the interface by scanning over the model to see if there exists dropout layers?

AnirudhDagar: I'd lean for not doing that. I've mentioned this before, removing and hiding all these JAX/Flax intricacies is not right, and if we are adding JAX/Flax then why not do it the right way. I don't understand why are we so against these, there is nothing wrong with the key argument, it is used everywhere in Flax. In fact, we need to show that it is there.

Even if we hide it, it will only make the JAX version slower by scanning over the model to see if there exists dropout layers.

AnirudhDagar: Ok, so based on our offline conversation, I've made a small hack to hide the key argument (making it optional).
`trainer.fit(model, data, key={'params': get_key(), 'dropout': get_key()})` -> `trainer.fit(model, data)`

We now pass the dropout key for all the models, and it will only be used by flax when a dropout layer is encountered. (not the most elegant, could also confuse a few readers. Hence I add a comment to explain).

I'm not a big fan of this approach about hiding the key, as i mentioned in our discussion, but it addresses your concerns about hiding the key, and for the readers, it looks simpler/readable. Also code across frameworks looks similar as requested.

astonzhang: Technically, it is "nothing wrong" (make the code runnable) to have `trainer.fit(model, data, key)`. It is also "nothing wrong" to have something like `trainer.fit(model, data, num_layers)`. And it is also "nothing wrong" to keep adding arguments to `fit` as we keep having "special reasons" when growing the codebase. However, since `key` is the concern of `model`,  a more **scalable** design will recommend hiding `key` inside `model`.

I'm glad you found where it should go: yes, `loss` is part of the model :) I agree that there may be better ways, but given the cleaner (and more uniform) front-end in this version, let's stick to it for now and keep an eye on future improvement opportunities.

PR #2319, opened 2022-10-02
Fix typo

gab-chen:
Here both X and w are matrix. As defined in the code, w's shape is (num_inputs, 1). Xw is a matrix-matrix product, instead of a matrix-vector product.  

For d2l.matmul(X, self.w), a matrix-vector product returns a 1-d vector, while a matrix-matrix product returns a multi-dimension matrix.  Thus it is important to fix the typo because it is confusing.

*Description of changes:*

choice.


astonzhang (commented):
In math, w is a vector. To make sure we are talking math when using math notation, I just added a line to refer to [Equation (3.1.4)](http://preview.d2l.ai.s3-website-us-west-2.amazonaws.com/d2l-en/master/chapter_linear-regression/linear-regression.html)

PR #2320, opened 2022-10-02
Remove unnecessary .reshape()

gab-chen:
*Description of changes:*
The .reshape() here is unnecessary, because both y_hat and y have the same shape (32, 1). 

The additional .reshape() is confusing to the beginners.  
When I started to read the book, I struggled with this .reshape() and spent a lot of time trying to understand why it is needed. But it is actually not needed. I hope the future beginners do not have to waste their time on this. 

choice.

### DIFF
--- a/chapter_linear-regression/linear-regression-scratch.md
+++ b/chapter_linear-regression/linear-regression-scratch.md
@@ -130,7 +130,7 @@ among all examples in the minibatch.
 %%tab all
 @d2l.add_to_class(LinearRegressionScratch)  #@save
 def loss(self, y_hat, y):
-    l = (y_hat - d2l.reshape(y, y_hat.shape)) ** 2 / 2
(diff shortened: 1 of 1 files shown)


AnirudhDagar (approved):
Thanks @gab-chen!

IIRC, previously, we used the squared loss in chapters later, which required this extra step of reshaping. reshape here is rightly not required anymore since we are localizing the loss for different models with the new lightning like d2l api.

I'm +1 on the change.

AnirudhDagar:
@gab-chen Could you please send another PR to replace your id with your name in our [acknowledgement](https://github.com/d2l-ai/d2l-en/blob/master/chapter_preface/index.md)?

gab-chen:
@AnirudhDagar Thanks for reviewing the pull request. 

If it is ok, I will leave my Github id as it is because I feel it would provide more information about me.

PR #2324, opened 2022-10-03
JAX: Add lenet.md & switch to channels last fashionmnist

AnirudhDagar:
This PR makes some major changes:

2. Update init method to add support for GPUs during training. Note that jax raises an error if gpu backend is not found when using `jax.device_count('gpu')`. Hence we switch to `jax.device_count()` without specifying the backend, but this returns the gpu/xla backends plus the CPU. Hence we subtract one, to get the true accelerated device count without running into runtime errors on machines which don't have any GPU.
3. Add lenet section for JAX.


AnirudhDagar:
@astonzhang made the changes you requested regarding not using `tensorflow_datasets`/`tensorflow`, this is now ready for review.

PR #2329, opened 2022-10-13
typo and timer usage fix 

regob:
*Description of changes:*
Typo: g(0) = f(y) from the definition (and not g(y))
Fixed timer usages: timer.avg() does not measure **time per epoch** if the code above stops/restarts
the timer during the epoch, because multiple numbers are added per epoch into the timer's list.

choice.


AnirudhDagar (approved):
Thanks @regob that makes sense. Indeed the current version measures the per epoch time incorrectly.

PR #2332, opened 2022-10-19
Update deep-rnn.md

LALBJ:
*Description of changes:*

I will received error `IndexError: list index out of range` while using the previous code.
Because the shape of inputs is [1, 1, 28].

### DIFF
--- a/chapter_recurrent-modern/deep-rnn.md
+++ b/chapter_recurrent-modern/deep-rnn.md
@@ -153,7 +153,7 @@ layer by layer.
 @d2l.add_to_class(StackedRNNScratch)
 def forward(self, inputs, Hs=None):
     outputs = inputs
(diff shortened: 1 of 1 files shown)


chapter_recurrent-modern/deep-rnn.md
```diff
@@ -153,7 +153,7 @@ layer by layer.
 @d2l.add_to_class(StackedRNNScratch)
 def forward(self, inputs, Hs=None):
     outputs = inputs
-    if Hs is None: Hs = [None] * len(inputs)
+    if Hs is None: Hs = [None] * inputs.shape[-1]
```
astonzhang, line 156:
```suggestion
    if Hs is None: Hs = [None] * self.num_layers
```

PR #2366, opened 2022-12-01
JAX: Add LSTM

AnirudhDagar:
*Description of changes:*
Add LSTM implementation in JAX. Also, remove the saved class `LSTMScratch` since it is not used in any of the other sections in d2l; there is no point in saving the class.

choice.

### DIFF
--- a/chapter_recurrent-modern/lstm.md
+++ b/chapter_recurrent-modern/lstm.md
@@ -318,13 +318,13 @@ Note that only the hidden state is passed to the output layer.
 [**The actual model**] is defined as described above,
 consisting of three gates and an input node. 
 Note that only the hidden state is passed to the output layer.
-A long `for` loop in the forward method will result in extremely long
-JIT compilation times for the first run. As a solution to this, instead
-of using a `for` loop to update the state with every time step,
-JAX has `jax.lax.scan` utility transformation to achieve the same behaviour.
+A long for-loop in the `forward` method will result in an extremely long
+JIT compilation time for the first run. As a solution to this, instead
+of using a for-loop to update the state with every time step,
+JAX has `jax.lax.scan` utility transformation to achieve the same behavior.
 It takes in an initial state called `carry` and an `inputs` array which
-is scanned on its leading axis. `scan` ultimately returns the final state
-and the stacked outputs as expected.
+is scanned on its leading axis. The `scan` transformation ultimately
+returns the final state and the stacked outputs as expected.
 :end_tab:
 
 ```{.python .input}
@@ -354,8 +354,8 @@ def forward(self, inputs, H_C=None):
 %%tab jax
 @d2l.add_to_class(LSTMScratch)
 def forward(self, inputs, H_C=None):
-    # Use lax.scan primitive instead of looping over the inputs
-    # scan saves time in jit compilation
+    # Use lax.scan primitive instead of looping over the
+    # inputs, since scan saves time in jit compilation.
     def scan_fn(carry, X):
         H, C = carry
         I = d2l.sigmoid(d2l.matmul(X, self.W_xi) + (
@@ -372,7 +372,7 @@ def forward(self, inputs, H_C=None):
 
     if H_C is None:
         batch_size = inputs.shape[1]
-        carry = jnp.zeros((batch_size, self.num_hiddens)),\
+        carry = jnp.zeros((batch_size, self.num_hiddens)), \
                 jnp.zeros((batch_size, self.num_hiddens))
     else:
         carry = H_C
--- a/chapter_recurrent-modern/lstm.md
+++ b/chapter_recurrent-modern/lstm.md
@@ -245,7 +245,6 @@ from d2l import jax as d2l
 from flax import linen as nn
 import jax
 from jax import numpy as jnp
-from functools import partial
 ```
 
 ### [**Initializing Model Parameters**]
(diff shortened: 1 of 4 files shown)


chapter_recurrent-modern/lstm.md
```diff
@@ -277,12 +285,50 @@ class LSTMScratch(d2l.Module):  #@save
...
+consisting of three gates and an input node. 
+Note that only the hidden state is passed to the output layer.
+:end_tab:
+
+
+:begin_tab:`jax`
 [**The actual model**] is defined as described above,
 consisting of three gates and an input node. 
 Note that only the hidden state is passed to the output layer.
+A long `for` loop in the forward method will result in extremely long
```
astonzhang, line 321:
```suggestion
A long `for` loop in the `forward` method will result in extremely long
```

chapter_recurrent-modern/lstm.md
```diff
@@ -277,12 +285,50 @@ class LSTMScratch(d2l.Module):  #@save
...
+JIT compilation times for the first run. As a solution to this, instead
```
astonzhang, line 322:
```suggestion
JIT compilation time for the first run. As a solution to this, instead
```

chapter_recurrent-modern/lstm.md
```diff
@@ -277,12 +285,50 @@ class LSTMScratch(d2l.Module):  #@save
...
+of using a `for` loop to update the state with every time step,
```
astonzhang, line 323:
```suggestion
of using a for-loop to update the state with every time step,
```

chapter_recurrent-modern/lstm.md
```diff
@@ -277,12 +285,50 @@ class LSTMScratch(d2l.Module):  #@save
...
+JAX has `jax.lax.scan` utility transformation to achieve the same behaviour.
```
astonzhang, line 324:
```suggestion
JAX has `jax.lax.scan` utility transformation to achieve the same behavior.
```

chapter_recurrent-modern/lstm.md
```diff
@@ -277,12 +285,50 @@ class LSTMScratch(d2l.Module):  #@save
...
+It takes in an initial state called `carry` and an `inputs` array which
+is scanned on its leading axis. `scan` ultimately returns the final state
```
astonzhang, line 326:
```suggestion
is scanned on its leading axis. The `scan` transformation ultimately returns the final state
```

astonzhang: Otherwise `scan` starts a sentence with a lower case.

chapter_recurrent-modern/lstm.md
```diff
@@ -304,14 +350,46 @@ def forward(self, inputs, H_C=None):
     return outputs, (H, C)
 ```
 
+```{.python .input}
+%%tab jax
+@d2l.add_to_class(LSTMScratch)
+def forward(self, inputs, H_C=None):
+    # Use lax.scan primitive instead of looping over the inputs
+    # scan saves time in jit compilation
```
astonzhang, line 358: Merge two comments into one single paragraph comment

chapter_recurrent-modern/lstm.md
```diff
@@ -304,14 +350,46 @@ def forward(self, inputs, H_C=None):
...
+                        d2l.matmul(H, self.W_ho) + self.b_o)
+        C_tilde = d2l.tanh(d2l.matmul(X, self.W_xc) +
+                           d2l.matmul(H, self.W_hc) + self.b_c)
+        C = F * C + I * C_tilde
+        H = O * d2l.tanh(C)
+        return (H, C), H  # return carry, y
+
+    if H_C is None:
+        batch_size = inputs.shape[1]
+        carry = jnp.zeros((batch_size, self.num_hiddens)),\
```
astonzhang, line 375:
```suggestion
        carry = jnp.zeros((batch_size, self.num_hiddens)), \
```

chapter_recurrent-modern/lstm.md
```diff
@@ -249,8 +257,8 @@ with 0.01 standard deviation,
 and we set the biases to 0.
 
 ```{.python .input}
-%%tab all
-class LSTMScratch(d2l.Module):  #@save
```
astonzhang, line 253: Add #@save back and re-run `d2lbook build lib`

d2l/mxnet.py
```diff
@@ -764,21 +764,6 @@ def init_params(self):
     def output_layer(self, hiddens):
         return d2l.swapaxes(self.linear(hiddens), 0, 1)
 
-class LSTMScratch(d2l.Module):
```
astonzhang, line 767: Should we add back `LSTMScratch` to d2l lib for all the frameworks?

astonzhang:
LGTM after the comments are addressed.

AnirudhDagar:
The implementation I have in JAX here is already written in a way that those comments from #2314 are satisfied and do not apply. We always initialize the state before the `for-loop` aka `scan` in JAX. So it doesn't affect JAX. At the merge time for JAX to master, this should be pretty straightforward because the changes will not force us to change JAX implementation.

I'll make the other suggested changes and update the PR, but I'd strongly suggest avoiding any change in the JAX branch right now. I'll be handling the merge conflicts later, anyway, so it is easiest to handle those at once.

PR #2433, opened 2023-02-01
Remove unnecessary reshape

gab-chen:
*Description of changes:*
Please let me know if you need more details. 
I tested removing the `.reshape((-1, 1))` from the PyTorch version of the notebook. The train loss and validation loss remain the same after the removal. But I'm unsure how the code changed here translates to other framework versions like JAX or TF. Can you double check if the change will work for non-PyTorch versions?

Removal of the unnecessary `reshape` helps ease the burden on the learners at this early chapter of the book. 

choice.


gab-chen:
I skimmed through the error trace. The errors occurred at .jax files. Does that mean the change is incompatible with JAX only?

astonzhang:
Thanks. At this point we'll keep it for compatibility with other frameworks with shared code cells.

PR #2434, opened 2023-02-01
Remove unnecessary reshape

gab-chen:
*Description of changes:*
Please let me know if you need more details.
I tested removing the .reshape((-1, 1)) from the PyTorch version of the notebook. The `multistep_preds` plot remains the same. 

choice.


gab-chen:
I checked the error trace. The errors occurred at files under `site-packages/mxnet/`. Does that mean the change is incompatible with MXNET only? If so, how can I update the code to make it compatible with MXNET version? If it is too much work, I can just close the pull request.

astonzhang:
Thanks. At this point we'll keep it for compatibility with MXNet with shared code cells.

PR #2441, opened 2023-02-15
Update index.md

frizfealer:
*Description of changes:*

Update the d2l version to be the same as the one in Chinese version.

choice.


astonzhang:
The English version uses a different package version

PR #2468, opened 2023-04-14
Update linear-algebra.md

initzx:
The matrix multiplication operator is not due to PyTorch, but due to https://peps.python.org/pep-0465. PyTorch's tensor classes simply overrides the __matmul__ method. 

Best regards

*Description of changes:*

choice.


astonzhang:
Good point!

PR #2475, opened 2023-04-28
Update W matrix subscript to d

rajrkane:
The weight matrix is said to be an element of $\mathbb{R}^{d \times h}$, but is denoted as $\mathbf{W}_{xh}$ with the first subscript $x$ throughout the page. $x$ is nowhere defined, so this seems to be a typo.  The proposed changes replace the subscript $x$ with $d$ throughout the page where written in math notation or in code snippets, thus making the notation mathematically correct and consistent.

*Description of changes:*

choice.


RohanKarthikeyan:
I think that it is right to leave it as $\textbf{W}_{xh}$ because it denotes the weights "between" the inputs (which you may denote by $x$) and the hidden layers (which you may denote by $h$). 

The subscripts under the weights $\textbf{W}$, I feel, do not intend to convey the dimensionality of the matrices.

But the authors are free to comment on this.

astonzhang:
Thanks. Our notation is for consistency with those in chapter_recurrent-modern/gru.html

PR #2484, opened 2023-05-09
CI: Exclude s3 data upload & Add utils.sh workflow script

AnirudhDagar:
This PR
* improves S3 cache sync time by excluding data file upload. [10min saved on PyTorch; 13min saved on MXNet S3 uploads]
* Add a util function to measure command time for S3 cache sync [Works]
* test `clear_cache` functionality through a single flag in utils.sh [Works]
* test maintainers' ability to change workflow_script files through PRs (without a direct push) [Works]

### DIFF
--- a/.github/workflow_scripts/utils.sh
+++ b/.github/workflow_scripts/utils.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # Default value for CLEAR_CACHE
-CLEAR_CACHE=true  # Eg. 'true' or 'false'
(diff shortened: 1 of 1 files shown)


AnirudhDagar:
Note to self: Retrieving the commit message from PRs securely is not that straightfwd. One will have to fetch from PRs source ref. I'll look into the implementation of clear cache later.

AnirudhDagar:
Instead of using commit-msg to clear the cache, maintainers may disable it by changing the flag in the workflow script `utils.sh` for a PR.

PR #2485, opened 2023-05-10
Update vgg.md

cx-olquinjica:
variable reference error. It should be _arch_ instead of conv_arch to define the variable containing the tuple (num_convs, out_channels). 

Change variable name: conv_arch > arch

*Description of changes:*

choice.


AnirudhDagar (approved):
Thanks @cx-olquinjica, LGTM!

PR #2492, opened 2023-05-15
Add little description of dataset downloading by torchvision.datasets.XXDATASET(xx,download=True)

zzowenzz:
*Description of changes:*

choice.

### DIFF
--- a/chapter_builders-guide/model-construction.md
+++ b/chapter_builders-guide/model-construction.md
@@ -108,6 +108,7 @@ npx.set_np()
 import torch
 from torch import nn
 from torch.nn import functional as F
+import torchinfo
 ```
 
 ```{.python .input}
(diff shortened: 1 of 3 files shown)


zzowenzz:
Hi maintainers, I was curious about using torchvision.datasets function to download dataset before but never really find out the file format and reason of the file downloaded. Add 1 paragraph to explain the file format (IDX) is to make it understandable for other readers.

AnirudhDagar (commented):
@zzowenzz please consider making two different PRs for your commits; these changes are orthogonal. Using `torchviz` to visualize the computation graph is a nice idea (yes we'll have to add `graphviz`, before CI can pass), though it would be a PyTorch-specific thing. Not sure if we can do something similar for all the frameworks? @astonzhang what do you think?

zzowenzz:
Hi @AnirudhDagar, thanks for your kind suggestions. Will create a separate PR for the visualisation thing. And for the installation of that library, I first tried to install from the environment level using pip as I don't want to mess up the working environment. But it seems that this method failed as it encounter a PATH problem. I will try to test the installation for other frameworks.

AnirudhDagar:
@zzowenzz don't worry about graphviz installation atm, we can add that to the dockerfile, if we do decide to go ahead with the change. It would be nice to see a screenshot of the viz anyway in the PR, even if CI fails.

zzowenzz:
@AnirudhDagar Sure. Example code as
```
import torch
from torchviz import make_dot

# Create some tensors and perform operations
x = torch.tensor([2.0], requires_grad=True)
y = torch.tensor([3.0], requires_grad=True)
z = x + y
output = z * z

# Visualise the computational graph
dot = make_dot(output, params={"x": x, "y": y})

# Save the graph as a PDF file
dot.render("computational_graph", format="png")
```
will generate an image which shows like this:
[computational graph](https://drive.google.com/file/d/1PWCXS2anXhMJM3XtZi6HNeGyjsTJB9Ex/view?usp=share_link)
which clearly demonstrates the nodes (like x, y), edges (with direction) and operation (like AddBackward0, MulBackward0)

zzowenzz:
Will create separate PRs

PR #2500, opened 2023-05-25
Update lstm.md

cx-olquinjica:
Rearrange the order of the words to improve clarity

Original: "Hochreiter had articulated this problem __as early as in his 1991 masters thesis__, although the results were not widely known because the thesis was written in German. "

Changed to: "Hochreiter had articulated this problem __as early as 1991 in his masters thesis__, although the results were not widely known because the thesis was written in German. "

*Description of changes:*

choice.


AnirudhDagar (approved):
Thanks @cx-olquinjica; this reads better to me as well.

PR #2541, opened 2023-08-10
CI: Update jax and tensorflow containers

AnirudhDagar:
This PR adds support for the latest version of JAX and TF containers for D2L.


astonzhang:
Can you also update versions in https://github.com/d2l-ai/d2l-en/blob/master/chapter_installation/index.md ?

PR #2613, opened 2024-07-19
fix: Update setup.py to support Python 3.12.4 and NumPy 2.0.0.

slewsys:
Update setup.py for Python 3.12.4 and NumPy 2.0.0.  
This PR is compatible with the latest iterations of Python v3.10, v3.11 and v3.12.  Python v3.9 is not supported.

choice.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,13 +2,13 @@ from setuptools import setup, find_packages
 import d2l
 
 requirements = [
-    'jupyter==1.0.0',
-    'numpy==1.23.5',
-    'matplotlib==3.7.2',
-    'matplotlib-inline==0.1.6',
-    'requests==2.31.0',
-    'pandas==2.0.3',
-    'scipy==1.10.1'
+    'jupyterlab==4.2.2',
+    'numpy==2.0.0',
+    'matplotlib==3.9.1',
+    'matplotlib-inline==0.1.7',
+    'requests==2.32.3',
+    'pandas==2.2.2',
+    'scipy==1.14.0'
 ]
 
 setup(


RahulVadisetty91 (commented):
image

In the requirements missing comma here correct it now. by the way all are good

setup.py
```diff
@@ -2,13 +2,13 @@
...
-    'jupyter==1.0.0',
-    'numpy==1.23.5',
-    'matplotlib==3.7.2',
-    'matplotlib-inline==0.1.6',
-    'requests==2.31.0',
-    'pandas==2.0.3',
-    'scipy==1.10.1'
+    'jupyterlab==4.2.2',
+    'numpy==2.0.0',
+    'matplotlib==3.9.1',
```
aliphys, line 7: 3.9.1 is yanked and fails to build on Windows. 
```suggestion
    'matplotlib==3.9.2',
```

```
warning: `matplotlib==3.9.1` is yanked (reason: "The Windows wheels, under some conditions, caused segfaults in unrelated user code.  Due to this we deleted the Windows wheels to prevent these segfaults, however this caused greater disruption as pip then began to try (and fail) to build 3.9.1 from the sdist on Windows which impacted far more users.  Yanking the whole release is the only tool available to eliminate these failures without changes to on the user side.  The sdist, OSX wheel, and manylinux wheels are all functional and there are no critical bugs in the release.   Downstream packagers should not yank their builds of Matplotlib 3.9.1.  See https://github.com/matplotlib/matplotlib/issues/28551 for details.")
  × Failed to build `matplotlib==3.9.1`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `mesonpy.build_wheel` failed (exit code: 1)                                                                                                                                                                                                                                                                                                         
```

PR #2676, opened 2026-01-02
Update oo-design.md

abhiqaz18:
*Description of changes:*

Fixed assertion failure message. This also needs to be changed in the source of the d2l.Module class. I'll submit a separate pull request for that.

choice.


carlaarchuleta308-source (changes requested):
abhiqaz18:patch-1

PR #2702, opened 2026-04-20
Fix finite-difference example in conv-layer to use horizontal neighbors

### LINKED ISSUE #2669: conv-layer.ipynb
"Note that this kernel is a special case of a finite difference operator. At location  (𝑖,𝑗)  it computes  𝑥𝑖,𝑗−𝑥(𝑖+1),𝑗 , i.e., it computes the difference between the values of horizontally adjacent pixels. "

For this example of the finite difference operation, I think it computes x i,j - x i,j+1 instead.

Chessing234:
Closes #2669.

## Bug

The paragraph introducing the `[1, -1]` edge-detection kernel in `chapter_convolutional-neural-networks/conv-layer.md` says the kernel

> computes the difference between the values of horizontally adjacent pixels [and] is a discrete approximation of the first derivative in the horizontal direction

but the two formulas on that same line describe a *vertical* difference:

```
x_{i,j} - x_{(i+1),j}
-\partial_i f(i,j) = lim (f(i,j) - f(i+eps,j)) / eps
```

## Root cause

With the book's (row `i`, column `j`) indexing, moving in `j` is horizontal; moving in `i` is vertical. The kernel `K = [[1.0, -1.0]]` is 1×2 and applied along the width axis, so at position `(i, j)` cross-correlation produces

```
1 * X[i, j] + (-1) * X[i, j+1]  =  X[i, j] - X[i, j+1]
```

i.e. a difference between horizontal neighbors, approximating the partial derivative along `j`. The prose is correct; the two symbolic expressions used the wrong axis.

## Why the fix is correct

Three edits on that single line, all swapping the `i`-axis offsets for `j`-axis ones so the formulas match the surrounding prose and the actual behaviour of the 1×2 kernel:

- `x_{i,j} - x_{(i+1),j}` → `x_{i,j} - x_{i,j+1}`
- `-\partial_i f(i,j)` → `-\partial_j f(i,j)`
- `f(i+\epsilon, j)` → `f(i, j+\epsilon)`

No other changes; the Python block directly below (`K = d2l.tensor([[1.0, -1.0]])` and the worked example on a 6×8 input) is unaffected and already illustrates the horizontal interpretation.


smolix:
Hi @Chessing234 - I'm in the middle of refactoring d2l to move it to a more modern tooling (Quarto, latest versions of JAX, TF/Keras and PyTorch, better references, fix performance bugs, etc.). I'll have a first cut of this shortly. Reach out to me by e-mail and we can discuss more (no idea who you are, based off your GH info).

Chessing234:
Thanks for the heads up! I will reach out to you by email to discuss this further. Closing this PR for now given the upcoming refactor.

Issue #2734, opened 2026-08-30
Add a readable original-paper link to the BERT chapter?


euwyngoh:
The current BERT chapter cites Devlin et al. through the book bibliography, but learners do not get a direct route from the chapter to the complete paper.

Web Editions has a section-addressable HTML reader with the original ACL source attached:

https://webeditions.page/works/bert-pre-training/

Would it be useful to add this as `Read the original paper online` near the Devlin citation or in Further Reading? It is meant as a mobile companion, not a replacement for the source of record. Happy to send a small PR if that fits the book's link policy.

PR #2735, opened 2026-09-05
Link a readable version of the original BERT paper

euwyngoh:
*Description of changes:*

Adds a direct section-addressable reading link to the original BERT paper beside its existing citation. The book's citation remains unchanged.

Relates to #2734.

choice.


euwyngoh:
CI note: all four jobs reached the same upstream infrastructure failure after four hours: the AWS Batch credential expired while the workflow was still polling the queued job (ExpiredTokenException). The proposed change itself is limited to two Markdown lines beside the existing BERT citation.

euwyngoh:
Hi, a quick follow-up because all four checks failed on the same expired AWS token rather than this two-line Markdown change. Would you be open to rerunning CI or merging the link update as-is? Happy to close it if optional reader links are not a fit for D2L.

